PR #5, opened 2020-11-06
Adds shortcut script to start sever

dsmith42:
To make it easier to start our local server have added an executable bash script to start the server to the root directory.
To fire it up just type in:
`./start-server.sh`


Gmiller290488 (commented):
Really nice.
Would you mind just editing the ReadMe to go along with this change please?

Gmiller290488:
You know I was thinking on this a bit and we could take it a stage further...
Create a runscript that would run when the project builds.
We could check if the port is in use (in which case we know the server is running and do nothing), otherwise spin up the server. Not sure how we could gracefully handle the case when `json-server` isn't installed. I guess we could `echo You need to run 'npm install -g json-server'` or something? 🤷 
Let me know what you think :-)

dsmith42:
> You know I was thinking on this a bit and we could take it a stage further...
> Create a runscript that would run when the project builds.
> We could check if the port is in use (in which case we know the server is running and do nothing), otherwise spin up the server. Not sure how we could gracefully handle the case when `json-server` isn't installed. I guess we could `echo You need to run 'npm install -g json-server'` or something? 🤷
> Let me know what you think :-)

I think that is a nice idea, it would make things more seamless.  I'll take a look at that this morning.

raynelle03 (approved):

PR #6, opened 2020-11-06
Feature/adds UUID generator

dsmith42:
Found a useful package for KMP that should enable us to generate new UUIDs that work on iOS and Android. [uuid](https://github.com/benasher44/uuid)

Have added it to the project and implemented an example usage.

### DIFF
--- a/iosHackthonApp/iosHackthonApp/ContentView.swift
+++ b/iosHackthonApp/iosHackthonApp/ContentView.swift
@@ -55,13 +55,8 @@ extension ContentView {
                     self.meals = .error(error?.localizedDescription ?? "error")
                 }
             })
-            self.getNewUUID()
-            self.getNewUUID()
-            self.getNewUUID()
-        }
-
-        func getNewUUID() {
-          print("UUID test: \(sdk.getUUID())")
+            print("UUID example call: \(sdk.getUUID())")
+            print("UUID example call: \(sdk.getUUID())")
         }
     }
 }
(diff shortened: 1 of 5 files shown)


iosHackthonApp/iosHackthonApp/ContentView.swift
```diff
@@ -55,6 +55,13 @@ extension ContentView {
                     self.meals = .error(error?.localizedDescription ?? "error")
                 }
             })
+            self.getNewUUID()
```
raynelle03: Just wondering..why are we calling it 3 times?

shared/build.gradle.kts
```diff
@@ -28,7 +28,7 @@ kotlin {
     val coroutinesVersion = "1.3.9-native-mt"
     val serializationVersion = "1.0.0-RC"
     val ktorVersion = "1.4.0"
-    val touchlabVersion = "0.2.5"
```
Gmiller290488, line 31: 👍

Gmiller290488 (approved):

iosHackthonApp/iosHackthonApp/ContentView.swift
```diff
@@ -55,6 +55,13 @@ extension ContentView {
...
+            self.getNewUUID()
+            self.getNewUUID()
+        }
+
+        func getNewUUID() {
+          print("UUID test: \(sdk.getUUID())")
```
Gmiller290488: I like the idea of showing that it works, but let's remove the calls to `getNewUUID()` :-)

raynelle03 (approved):

iosHackthonApp/iosHackthonApp/ContentView.swift
dsmith42: I just wanted to check that it gave a different UUID each time.   Will remove additional calls now.

PR #10, opened 2020-11-09
Fixes build failure due to missing self in closure

dsmith42:
I was getting build failures for the project.  When I opened it in Xcode it had an error that `self` was required inside a closure so have rectified.


dsmith42 (commented):
Looks fine

raynelle03 (approved):

PR #13, opened 2020-11-10
Showing collection code after POST meal

Gmiller290488:
I added code to present an alert displaying a collection code if a Meal is successfully added.
Otherwise show an error message.


iosHackthonApp/iosHackthonApp/Add Meal Screen/AddMealView.swift
```diff
@@ -11,45 +11,45 @@ import shared
...
-                quantity: 2,
-                availableFromDate: "Tuesday",
-                expiryDate: "Saturday",
-                info: "Meat",
-                hot: false,
-                locationLat: 51.509865,
-                locationLong:  -0.118092)
-            self.viewModel.postMeal(meal: meal)
+        ZStack {}
+        .alert(isPresented: $viewModel.showingCollectionCode) {
```
Gmiller290488, line 18: SwiftUI gets upset if you have two alerts on the same view so this is just here temporarily.
When we get to the stage of showing something more elaborate, we can then get rid of this ZStack and alert code.

iosHackthonApp/iosHackthonAppTests/Strings+extensionTests.swift
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    func test_last4Chars_asNumbers_returnsCorrectString() {
+        XCTAssertEqual("abcg1212".last4Chars(), "1212")
+    }
+
+    func test_last4Chars_asCombined_returnsCorrectString() {
+        XCTAssertEqual("abcg12".last4Chars(), "CG12")
+    }
+}
```
dsmith42, line 25: Nice - we have some tests 😀

dsmith42 (approved):
Looks good to me👍

iosHackthonApp/iosHackthonAppTests/Strings+extensionTests.swift
Gmiller290488: Only a few 🙈

PR #17, opened 2020-11-12
ListView UI

Gmiller290488:
Quite a large PR but a lot of the files are just indentation changes and adding images


iosHackthonApp/iosHackthonApp.xcodeproj/project.pbxproj
```diff
@@ -546,7 +566,7 @@
 				GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE;
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				GCC_WARN_UNUSED_VARIABLE = YES;
-				IPHONEOS_DEPLOYMENT_TARGET = 13.2;
+				IPHONEOS_DEPLOYMENT_TARGET = 14.0;
```
Gmiller290488, line 597: Bumped this because using LazyVstack is only available in ios14 and with a List view I couldn't get rid of the separators

iosHackthonApp/iosHackthonApp/Location/LocationManager.swift
```diff
@@ -12,38 +12,44 @@ import CoreLocation
...
-	override init() {
-		super.init()
-		self.locationManager.delegate = self
-		self.locationManager.desiredAccuracy = kCLLocationAccuracyBest
-		self.locationManager.requestWhenInUseAuthorization()
-		self.locationManager.startUpdatingLocation()
-	}
+    @Published var userLatitude: Double = 0
+    @Published var userLongitude: Double = 0
+    @Published var address: String = ""
```
kirstybutler, line 17: aren't the lat & long `Float`?

shared/src/commonMain/kotlin/com/kcc/kmmhackathon/shared/entity/Entity.kt
```diff
@@ -23,4 +23,17 @@ data class Meal(
     val locationLat: Float,
     @SerialName("locationLong")
     val locationLong: Float
+)
+
+@Serializable
+data class Quantity(
+    @SerialName("quantity")
+    val quantity: Int
+)
```
Gmiller290488, line 32: Added this in so that I can send just the quantity as the body for a patch request.

shared/src/commonMain/kotlin/com/kcc/kmmhackathon/shared/network/MealApi.kt
```diff
@@ -15,25 +16,30 @@ class MealApi {
             val json = Json { ignoreUnknownKeys = true }
             serializer = KotlinxSerializer(json)
             defaultRequest {
-                contentType(ContentType.Application.Json)
```
Gmiller290488, line 18: Was still having some issues with this as a default request 😖

iosHackthonApp/iosHackthonApp/Location/LocationManager.swift
roger-tan: I've seen in the past lat & long being a string instead of Double/Float to not lose the precision in their database.

Gmiller290488: Double is a double sized float :-)

iosHackthonApp/iosHackthonApp/Resources/Strings/Strings.swift
```diff
@@ -9,47 +9,90 @@
...
+            static let coldFood = "snow"
+            static let location = "location"
+        }
+
+        struct Date {
+            static let americanFormat = "yyyy-MM-dd' 'HH:mm:ssZ"
+            static let nameDayMonth = "E d MMM"
+        }
+    }
 }
```
kirstybutler, line 97: Maybe super picky but couldn't we have all the Alerts in once Struct? Or would that not work?

kirstybutler: Or all the errors together

shared/src/commonMain/kotlin/com/kcc/kmmhackathon/shared/network/MealApi.kt
kirstybutler: maybe we should leave the default request out then for what it's worth

iosHackthonApp/iosHackthonApp/Resources/Strings/Strings.swift
Gmiller290488: Yeah we could do.  I split them into the screens just because we have two collection alerts which are quite similar so wanted it to be obvious which one is for which screen

Gmiller290488: But having said that I would quite like to get rid of the collection alert and have that as a separate UI component.
When I do that, I can get rid of the screen specific alerts :-)

shared/src/commonMain/kotlin/com/kcc/kmmhackathon/shared/network/MealApi.kt
Gmiller290488: Yeah, we send the URL in each call anyway - I will get rid of it and check it still works as expected

iosHackthonApp/iosHackthonApp/Resources/Strings/Strings.swift
kirstybutler: sounds good 🙌

iosHackthonApp/iosHackthonApp/Location/LocationManager.swift
Gmiller290488: Interesting @roger-tan. These are just used locally for us to figure out how far away a meal is from the users current location :-)

kirstybutler (approved):

PR #22, opened 2020-11-12
Feature/add meal tweaks

raynelle03:


iosHackthonApp/iosHackthonApp/Components/ViewModifiers.swift
```diff
@@ -10,37 +10,45 @@ import Foundation
...
 struct GrayTextFieldStyle: ViewModifier {
-    func body(content: Content) -> some View {
-        content
-            .padding(.all)
-            .background(Color(red: 239.0/255.0, green: 243.0/255.0, blue: 244.0/255.0, opacity: 1.0))
-    }
+	func body(content: Content) -> some View {
+		content
+			.padding(.all)
+			.background(Color(red: 239.0/255.0, green: 243.0/255.0, blue: 244.0/255.0, opacity: 1.0))
```
Gmiller290488: Would be cool to add this to the assets and colorManager.
That way if we wanted to in the future have a dark mode, we can just add the equivalent dark code colour to that same asset and don't need to do anything else!
Let me know if you want a hand with it (if that ^ doesn't make sense 😅 )

Gmiller290488 (approved):
Just that one change for me please 👼

iosHackthonApp/iosHackthonApp/Components/ViewModifiers.swift
raynelle03: makes perfect sense..added this to assets now😀

PR #29, opened 2020-11-16
Testing mealEndpoint

Gmiller290488:
Changed how we get the correct endpoint string for android/ios...it seemed like a good case for the expect/actual rather than doing an if statement inside the API class.

Let me know if you think the other way was better :-)

### DIFF
--- a/shared/src/androidTest/kotlin/com/kcc/kmmhackathon/shared/network/MealEndpointTest.kt
+++ b/shared/src/androidTest/kotlin/com/kcc/kmmhackathon/shared/network/MealEndpointTest.kt
@@ -1,0 +1,12 @@
+package com.kcc.kmmhackathon.shared.network
+
+import kotlin.test.Test
+import kotlin.test.assertTrue
+
+class MealEndpointTestAND {
+
+    @Test
+    fun testMealEndpoint() {
+        assertTrue("10.0.2.2" in MealEndpoint().endpointString)
+    }
+}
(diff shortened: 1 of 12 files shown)


shared/src/androidTest/kotlin/com/kcc/kmmhackathon/shared/network/MealEndpointTest.kt
```diff
@@ -0,0 +1,12 @@
+package com.kcc.kmmhackathon.shared.network
+
+import kotlin.test.Test
+import kotlin.test.assertTrue
+
+class MealEndpointTestAND {
```
Gmiller290488, line 6: Have to differentiate the class name (hence the "AND" on the end) but having the filename be the same across CommonTest, AndroidTest, IOSTest isn't an issue

shared/src/commonMain/kotlin/com/kcc/kmmhackathon/shared/network/MealEndpoint.kt
```diff
@@ -0,0 +1,5 @@
+package com.kcc.kmmhackathon.shared.network
+
+expect class MealEndpoint() {
+    val endpointString: String
+}
```
kirstybutler, line 5: like the use of expect/actual :)

dsmith42 (approved):
Looks good

kirstybutler (approved):

PR #33, opened 2020-11-16
Added modules with Swift Package Manager

KaneCheshire:
This introduces modularisation to the iOS app, using Swift Package Manager packages instead of the traditional way of using multiple Xcode projects which is a total pain to manage.

On the contrary, module management is usually very simple with SPM because you only have to worry about the Package.swift file for each module. Since Xcode 12 we can include resources and assets in Packages which means that we can put all colours in a Theming module, for example.

Benefits of modules:
- It forces you to think about separation of concerns
- It forces you to think about what needs to be exposed publicly, which also leaves you free to leave some stuff exposed internally for testing
- It makes code more reusable, especially useful if you start adding iOS extensions and Mac apps etc
- It's tidy!
- It should improve build times, because Xcode (should) only rebuild modules that need rebuilding]

If this gets merged I can do the Android equivalent using Gradle :) 

Ideally, you have two types of modules, feature and shared (Android call them features and libs). Feature modules generally contain the screens for a single feature (usually a screen in the app), like MealList, and features shouldn't really ever know about other features. You can import libs into features, and libs into other libs if needed (like Theming).

Any questions, please shout! It's pretty epic that we can use a brand new Xcode integration (SPM) with a Kotlin framework.


iosHackthonApp/Modules/Features/MealList/Sources/MealList/MealListView.swift
```diff
@@ -1,12 +1,17 @@
...
+import Strings
 
-struct MealListView: View {
+public struct MealListView: View {
 
-    @ObservedObject private(set) var viewModel: MealListViewModel
+    @ObservedObject private var viewModel: MealListViewModel
 
-    var body: some View {
+	public init(viewModel: MealListViewModel) {
```
KaneCheshire: Ugh my indentation is off, I'll fix

iosHackthonApp/Modules/Features/MealList/Tests/LinuxMain.swift
```diff
@@ -0,0 +1,7 @@
+import XCTest
+
+import MealListTests
+
+var tests = [XCTestCaseEntry]()
```
KaneCheshire: This linux stuff isn't needed, unfortunately Xcode creates it automatically so I'll have to just delete

iosHackthonApp/Modules/Features/MealList/Tests/MealListTests/MealListTests.swift
```diff
@@ -0,0 +1,15 @@
+import XCTest
+@testable import MealList
+
+final class MealListTests: XCTestCase {
```
KaneCheshire, line 4: Need to add the tests to the scheme, but after that any new test functions are automatically added to the scheme just like any other test target

iosHackthonApp/Modules/Shared/Components/Package.swift
```diff
@@ -0,0 +1,30 @@
...
+    products: [
+        // Products define the executables and libraries a package produces, and make them visible to other packages.
+        .library(
+            name: "Components",
+            targets: ["Components"]),
+    ],
+    dependencies: [
+        // Dependencies declare other packages that this package depends on.
+        // .package(url: /* package url */, from: "1.0.0"),
+		.package(path: "../Shared/Theming")
```
KaneCheshire: Indentation

shared/build.gradle.kts
```diff
@@ -5,6 +5,7 @@ plugins {
     kotlin("plugin.serialization")
     id("com.android.library")
     id("kotlin-android-extensions")
+    id("com.chromaticnoise.multiplatform-swiftpackage") version "1.0.2"
```
KaneCheshire, line 8: This is the magic that lets us use SPM with a Kotlin framework, it's a third party gradle plugin that converts the build iOS framework into an XCFramework which is what SPM needs

KaneCheshire:
There's more to be done in terms of modularising, this only converts one feature (MealList) into a module and any code it depends on is broken out into modules too

iosHackthonApp/Modules/Features/MealList/Package.swift
```diff
@@ -0,0 +1,32 @@
...
+        .library(
+            name: "MealList",
+            targets: ["MealList"]),
+    ],
+    dependencies: [
+        // Dependencies declare other packages that this package depends on.
+		.package(path: "../Shared/Extensions"),
+		.package(path: "../Shared/Location"),
+		.package(path: "../Shared/Components"),
+		.package(name: "shared", path: "/Users/kanecheshire/Projects/benchHackathonNov20/shared/swiftpackage")
```
Gmiller290488: will this be a problem that the path is specific to your machine?

KaneCheshire: Oops! Yes, it's meant to be relative, I forgot to change. That explains why CI is broken!

iosHackthonApp/iosHackthonApp/Resources/Info.plist
```diff
@@ -67,7 +67,7 @@
 		<string>UIInterfaceOrientationLandscapeLeft</string>
 		<string>UIInterfaceOrientationLandscapeRight</string>
 	</array>
-    <key>UIUserInterfaceStyle</key>
-    <string>Light</string>
+	<key>UIUserInterfaceStyle</key>
```
KaneCheshire: No idea why this diff is showing, I'll try and revert

KaneCheshire:
Just CI to fix now

iosHackthonApp/Modules/Features/MealList/Package.swift
```diff
@@ -0,0 +1,32 @@
+// swift-tools-version:5.3
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
+import PackageDescription
+
+let package = Package(
```
KaneCheshire, line 5: This is all we need to manage a module really. The app target includes MealList as a target that gets build when the app does, which means the app can use this as a framework (anything that is exposed publicly), as well as any dependencies that MealList includes.

iosHackthonApp/Modules/Shared/Theming/Sources/Theming/ColorManager.swift
```diff
@@ -0,0 +1,19 @@
...
+import SwiftUI
+
+public struct ColorManager {
+
+    public static let red = Color.red
+    public static let blue = Color.blue
+    public static let gray = Color.gray
+	public static let appPrimary = Color("appPrimary", bundle: .module)
+    public static let appSecondary = Color("appSecondary", bundle: .module)
+	public static let textfieldGrey = Color("textField-grey", bundle: .module)
```
Gmiller290488: indentation is a bit off :-)

iosHackthonApp/Modules/Features/MealList/Package.swift
dsmith42: Niiice

shared/build.gradle.kts
dsmith42: So cool

aabolade: Alt Text

dsmith42 (approved):
Really nice, as you say so much better than having to make multiple projects

iosHackthonApp/Modules/Features/MealList/Package.swift
KaneCheshire: Dunno why I keep writing "build" instead of "built". I'm hoping it's just autocorrect lol

iosHackthonApp/Modules/Shared/Theming/Sources/Theming/ColorManager.swift
KaneCheshire: Thought I fixed it! I'll do it again

aabolade (approved):

.github/workflows/ios.yml
```diff
@@ -17,5 +17,7 @@ jobs:
         uses: actions/checkout@v2
       - name: Set Xcode version to 12.2
         run: sudo xcode-select -switch /Applications/Xcode_12.2.app
+      - name: Ensure Kotlin framework has built & converted to SPM
+        run: ./gradlew :shared:packForXCode -PXCODE_CONFIGURATION=${CONFIGURATION} && ./gradlew createSwiftPackage
```
KaneCheshire: Might not need this now

iosHackthonApp/Modules/Shared/Components/Package.swift
```diff
@@ -0,0 +1,30 @@
+// swift-tools-version:5.3
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
+import PackageDescription
+
+let package = Package(
+    name: "Components",
+    platforms: [.iOS(.v14)],
+    products: [
+        // Products define the executables and libraries a package produces, and make them visible to other packages.
```
KaneCheshire: don't need these comments

PR #34, opened 2020-11-16
Feature/android login ui

kirstybutler:
Really simple UI for now, pleaseee throw suggestions as we don't have any designs so just done the bare minimum.

Worth noting that the `imageView ` doesn't contain an image yet but is there as a placeholder for when we have a more suitable image (tried adding the icon we have at the min but just didn't sit right imo). So maybe one can be added at a later date.

Also added sections to some of the values files to make it easier to read.

Screenshot 2020-11-17 at 10 00 09


androidHackathonApp/src/main/res/values/dimens.xml
```diff
@@ -1,12 +1,23 @@
...
+    //Padding
     <dimen name="default_padding">16dp</dimen>
+
+    //Buttons
     <dimen name="button_image_size">50dp</dimen>
+    <dimen name="button_min_height">150dp</dimen>
+    <dimen name="button_radius">16dp</dimen>
+
+    //Edit Texts
+    <dimen name="edit_text_width">250dp</dimen>
```
kirstybutler, line 17: I'll remove this actually, this isn't used anymore

androidHackathonApp/src/main/res/values/dimens.xml
```diff
@@ -1,12 +1,23 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
+    //Margins
```
KaneCheshire, line 3: I didn't realise you could do `//` in the xml, I thought it had to be the old `<!--` way

kirstybutler: I'm not actually sure if `<!--` is still standard or not, to be honest. But I've used `//` for ages and no one's said anything lol

androidHackathonApp/src/main/res/values/styles.xml
```diff
@@ -3,9 +3,7 @@
     <style name="AppTheme" parent="Theme.AppCompat.Light.DarkActionBar">
         <!-- Customize your theme here. -->
```
kirstybutler, line 4: @KaneCheshire it's still used here, so idk

androidHackathonApp/src/main/res/layout/login.xml
```diff
@@ -0,0 +1,103 @@
...
+        android:id="@+id/login_edit_fields_layout"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:gravity="center_horizontal"
+        android:orientation="vertical"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toBottomOf="@id/login_image_view"
+        app:layout_constraintTop_toTopOf="parent">
```
KaneCheshire, line 28: Hmm is this right? You want to constrain to the top of parent _and_ the bottom of the image view?

androidHackathonApp/src/main/res/layout/login.xml
```diff
@@ -0,0 +1,103 @@
...
+            android:textSize="@dimen/default_text_size" />
+
+        <TextView
+            android:id="@+id/login_password_text"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_gravity="start"
+            android:layout_marginTop="@dimen/default_margin"
+            android:text="@string/title_password"
+            android:textSize="@dimen/default_text_size"
```
KaneCheshire: Probably better if we could use the text styles (appearances?) instead of harcoded text sizes, that way I think it changes for accessibility?

androidHackathonApp/src/main/res/values/dimens.xml
KaneCheshire: Much prefer //!

androidHackathonApp/src/main/res/layout/login.xml
kirstybutler: Oops! Nope - thanks for spotting

androidHackathonApp/src/main/res/layout/login.xml
kirstybutler: Nice! TIL

KaneCheshire:
I'm surprised CI passed with these deleted tbh
image

KaneCheshire (approved):

Gmiller290488 (approved):

PR #35, opened 2020-11-16
Added modules for Android

KaneCheshire:
Similar to the PR I created for introducing SPM modules to the iOS app, this PR adds modules using Gradle to the Android app.

Since the Android app is smaller it's a bit less of a faff to do, plus it's Kotlin importing Kotlin so much simpler. This PR breaks out the FindMealActivity into its own module, and also a Theming module to include the colours in.


androidHackathonApp/src/main/AndroidManifest.xml
```diff
@@ -18,7 +18,7 @@
             </intent-filter>
         </activity>
         <activity
```
KaneCheshire, line 20: Can probably move this into the manifest for the module because it merges them automatically

lib-find-meal/build.gradle.kts
```diff
@@ -0,0 +1,53 @@
+plugins {
```
KaneCheshire, line 1: There's a lot of repeated stuff in these gradle files, I can create a gradle file that has all the common stuff like compileSdkVersion etc so we can just include that file in each module

lib-find-meal/proguard-rules.pro
```diff
@@ -0,0 +1,21 @@
+# Add project specific ProGuard rules here.
```
KaneCheshire, line 1: Don't ask me what this file is for because I'm not experienced enough to know yet but I just leave them as-is for now!

lib-find-meal/src/androidTest/java/com/kinandcarta/lib/find/meal/ExampleInstrumentedTest.kt
```diff
@@ -0,0 +1,24 @@
+package com.kinandcarta.lib.find.meal
+
+import androidx.test.platform.app.InstrumentationRegistry
+import androidx.test.ext.junit.runners.AndroidJUnit4
+
```
KaneCheshire: I'll remove the example test files

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/FindMealActivity.kt
```diff
@@ -47,7 +45,7 @@ class FindMealActivity : AppCompatActivity() {
 
     override fun onDestroy() {
         super.onDestroy()
-        mainScope.cancel()
+        mainScope.cancel() // TODO: We'd get this for free in a ViewModel?
```
KaneCheshire: I'd really recommend using view models, because on Android they get lifecycle support for free

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/MealsAdapter.kt
```diff
@@ -13,11 +13,9 @@ import com.kcc.kmmhackathon.shared.entity.Meal
 
 class MealsAdapter (var mealsList: List<Meal>) : RecyclerView.Adapter<MealsAdapter.MealsViewHolder>() {
 
-    override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): MealsViewHolder {
-        return LayoutInflater.from(parent.context)
-            .inflate(R.layout.meal_item_row, parent, false)
-            .run(::MealsViewHolder)
```
KaneCheshire: So, when I moved the file this wouldn't compile and I had to change it. i'd like to change it back if anyone can figure out why it stopped compiling!

lib-find-meal/src/main/res/layout/find_meal_activity.xml
```diff
@@ -5,7 +5,7 @@
     xmlns:tools="http://schemas.android.com/tools"
     android:layout_width="match_parent"
     android:layout_height="match_parent"
-    tools:context=".FindMealActivity">
+    tools:context="com.kinandcarta.lib.find.meal.FindMealActivity">
```
KaneCheshire: Can probably be just `.FindMealActivity` again now

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/MealsAdapter.kt
kirstybutler: What was the error?

KaneCheshire: Dunno but it works now lol

lib-find-meal/build.gradle.kts
KaneCheshire: Struggling to get this working nicely with the .kts stuff

kirstybutler: When we did the tutorial to set up a skeleton project we had to refer to the many gradle files for different plugins/dependencies so I'm not sure how simple it would be to combine?

KaneCheshire: In a non-kts gradle file you can just apply from another .gradle file so it's easy to just put all the common dependencies etc in one file and just apply it, but I can't seem to get it to work with kts. I guess for now we can just repeat ourselves but would be nice to fix!

KaneCheshire: The blueprint does it (not with .kts extensions though) https://bitbucket.org/theappbusiness/android-blueprint/src/develop/buildScripts/common.gradle

kirstybutler (approved):

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/FindMealActivity.kt
kirstybutler: @dsmith42 @roger-tan are you both working on this? :)

PR #42, opened 2020-11-18
refactor add meal v2

Gmiller290488:


iosHackthonApp/Modules/Features/AddMeal/Sources/AddMeal/AddMealView.swift
```diff
@@ -143,8 +91,144 @@ public struct AddMealView: View {
     }
 }
 
+// MARK:- Views
+extension AddMealView {
```
KaneCheshire: Very minor but if you add `private` here you don't have to add it to the first-level nested types here because they're automatically private

Gmiller290488: Yeah sure :-) I'll do the same for the functions extension too 👍

iosHackthonApp/Modules/Features/AddMeal/Sources/AddMeal/AddMealView.swift
```diff
@@ -35,86 +41,28 @@ public struct AddMealView: View {
         title.isEmpty || additionalInfo.isEmpty || address.isEmpty || quantity == 0
     }
 
+    var buttonColor: Color {
```
raynelle03: can this be private?

iosHackthonApp/Modules/Features/AddMeal/Sources/AddMeal/AddMealView.swift
```diff
@@ -143,8 +91,143 @@ public struct AddMealView: View {
     }
 }
 
+// MARK:- Views
+private extension AddMealView {
+
+    struct InfoTextFieldsGroup: View {
+
+        @Binding var title: String
```
raynelle03, line 99: Can these and the rest of the variables be private?

Gmiller290488: I don't think these can be because the initialiser will match the access control of the properties, so if I make them private, the init becomes private; I have tried to get round it but haven't succeeded 😒 

But also the structs are private, and I don't keep a reference to the views anywhere, they are just created and used so I think it's alright

raynelle03 (approved):

KaneCheshire (approved):

aabolade (approved):

PR #43, opened 2020-11-18
Feature/AND location

dsmith42:
Adds getDistance to api and makes call from Android based on user location.  
I opted for creating a second api route for getMeals(userLat, userlLon, inMiles, forceReload) so that it has minimal impact on iOS, but should be able to update iOS frontend as well.

Sets up location permissions and will help us going forward into map view etc.  Massive activity at the moment, but hoping to work through it and refactor to ViewModel.

First time working with Android location stuff so do correct where things are unnecessary or done in a non kotlin / android best practice way 🙏


lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/view/FindMealActivity.kt
```diff
@@ -17,52 +27,148 @@ import kotlinx.coroutines.launch
...
+    private var locationUpdateState = false
+    private lateinit var lastLocation: Location
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
-        title = "Find a meal"
-        setContentView(R.layout.find_meal_activity)
+        setContentView(com.kinandcarta.lib.find.meal.R.layout.find_meal_activity)
+        title = getString(R.string.find_meal_title) // TODO figure out how to reach strings from main res/strings
```
kirstybutler: this todo can be removed now :)

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/view/FindMealActivity.kt
```diff
@@ -1,13 +1,23 @@
 package com.kinandcarta.lib.find.meal.view
```
kirstybutler, line 1: currently moving logic over to viewmodel for this

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/view/FindMealActivity.kt
```diff
@@ -17,52 +27,148 @@ import kotlinx.coroutines.launch
...
+    private val mealsAdapter = MealsAdapter(listOf())
 
     private lateinit var mealsRecyclerView: RecyclerView
     private lateinit var progressBarView: FrameLayout
     private lateinit var swipeRefreshLayout: SwipeRefreshLayout
 
-    private val sdk = MealsSDK()
-    private val mealsAdapter = MealsAdapter(listOf())
+    private lateinit var fusedLocationProviderClient: FusedLocationProviderClient
+    private lateinit var locationCallback: LocationCallback
```
KaneCheshire: I don't think this needs to be lateinit, I think you can just do
 
```kotlin
private val locationCallback = object : LocationCallback() {
... etc
}
```

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/view/FindMealActivity.kt
```diff
@@ -17,52 +27,148 @@ import kotlinx.coroutines.launch
...
+    private lateinit var locationRequest: LocationRequest
+    private var locationUpdateState = false
+    private lateinit var lastLocation: Location
```
KaneCheshire: Hmm this feels risky since we can't easily guarantee that lastLocation will be available before it's first used (basically the same as force-unwrapping a property in Swift). I'd probably just make this `var lastLocation: Location? = null`, however wondering if you can just get the last location directly from the client anyway?

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/view/FindMealActivity.kt
```diff
@@ -17,52 +27,148 @@ import kotlinx.coroutines.launch
...
+    private val sdk = MealsSDK()
+    private val mealsAdapter = MealsAdapter(listOf())
 
     private lateinit var mealsRecyclerView: RecyclerView
     private lateinit var progressBarView: FrameLayout
     private lateinit var swipeRefreshLayout: SwipeRefreshLayout
 
-    private val sdk = MealsSDK()
-    private val mealsAdapter = MealsAdapter(listOf())
+    private lateinit var fusedLocationProviderClient: FusedLocationProviderClient
```
KaneCheshire: If you want to avoid lateinit and make it immutable you can use `private val fusedLocationProviderClient: FusedLocationProviderClient by lazy { LocationServices.getFusedLocationProviderClient(this) }`

That's the same as doing `lazy var` in Swift and means it will only call getFusedLocationProviderClient when first accessed, and has the added bonus of making it immutable and not lateinit

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/view/FindMealActivity.kt
```diff
@@ -17,52 +27,148 @@ import kotlinx.coroutines.launch
...
+
+        fusedLocationProviderClient = LocationServices.getFusedLocationProviderClient(this)
+        updateLastLocation()
 
-        mealsRecyclerView = findViewById(R.id.rvMeals)
-        progressBarView = findViewById(R.id.progressBar)
-        swipeRefreshLayout = findViewById(R.id.swipeContainer)
+        locationCallback = object : LocationCallback() {
+            override fun onLocationResult(p0: LocationResult?) {
+                super.onLocationResult(p0)
```
KaneCheshire: I don't think you need to call super here

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/view/FindMealActivity.kt
```diff
@@ -17,52 +27,148 @@ import kotlinx.coroutines.launch
...
+
+                if (p0 != null) {
```
KaneCheshire: If you change `lastLocation` to be nullable then you can probably just do `lastLocation = locationResult?.lastLocation` without the need to check if locationResult is null (I'd also change the p0 arg name to be locationResult).

Alternatively I think you can do 
```kotlin
val location = locationResult?.lastLocation ?: return
lastLocation = location
```

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/view/FindMealActivity.kt
```diff
@@ -17,52 +27,148 @@ import kotlinx.coroutines.launch
...
+        }
+        displayMeals(false, false)
+    }
+
+    private fun updateLastLocation() {
+        if (ActivityCompat.checkSelfPermission(this, Manifest.permission.ACCESS_FINE_LOCATION) == PERMISSION_GRANTED
+        ) {
+            fusedLocationProviderClient.lastLocation.addOnSuccessListener { location: Location ->
+                if (location != null) {
+                    lastLocation = location
```
KaneCheshire: Same here RE the comment above about either using the elvis operator `?:` to get the location or just assign directly to lastLocation if lastLocation is nullable

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/view/FindMealActivity.kt
```diff
@@ -17,52 +27,148 @@ import kotlinx.coroutines.launch
...
+                    displayMeals(false, false)
```
KaneCheshire: This is my main gripe with Kotlin, without IDE help it's impossible to know what `false false` does here without finding the function and looking at the arg names

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/view/FindMealActivity.kt
```diff
@@ -17,52 +27,148 @@ import kotlinx.coroutines.launch
...
+    private fun startLocationUpdates() {
+        if (ActivityCompat.checkSelfPermission(this, android.Manifest.permission.ACCESS_FINE_LOCATION) != PERMISSION_GRANTED) {
+            ActivityCompat.requestPermissions(this, arrayOf(android.Manifest.permission.ACCESS_FINE_LOCATION), LOCATION_PERMISSION_REQUEST_CODE)
+            return
+        }
+        fusedLocationProviderClient.requestLocationUpdates(locationRequest, locationCallback, null)
+    }
+
+    private fun createLocationRequest() {
+        locationRequest = LocationRequest()
```
KaneCheshire: Does LocationRequest not have a Builder? 🤔 I thought it did

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/view/FindMealActivity.kt
```diff
@@ -17,52 +27,148 @@ import kotlinx.coroutines.launch
...
+    }
+
+    override fun onPause() {
+        super.onPause()
+        fusedLocationProviderClient.removeLocationUpdates((locationCallback))
+    }
+
+    override fun onResume() {
+        super.onResume()
+        if (!locationUpdateState) {
```
KaneCheshire, line 165: It would be nice not to need this check. Could we not just check location permissions each time we need to request location? That's what I do in my app. If location permissions are already granted then onPermissionsResult is called straight away and you can fire off the location request in there

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/view/FindMealActivity.kt
dsmith42: Nice, I like lazy :-)

gradle.properties
```diff
@@ -1,5 +1,6 @@
 kotlin.code.style=official
 xcodeproj=./iosHackthonApp
 android.useAndroidX=true
+android.enableJetifier=true
```
KaneCheshire, line 4: I'm curious, what does this do? 🤔 These are the bits of Droid dev I still really need to understand

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/adapter/MealsAdapter.kt
```diff
@@ -39,28 +39,28 @@ class MealsAdapter(var mealsList: List<Meal>) :
...
-            val tempColor = if (meal.hot) R.color.colorHot else R.color.colorCold
+        fun bindData(item: MealWithDistance) {
+            nameView.text = item.meal.name
+            val tempString = if (item.meal.hot) "Hot" else "Cold"
+            val tempColor = if (item.meal.hot) R.color.colorHot else R.color.colorCold
             tempView.text = tempString
             tempView.setTextColor(ContextCompat.getColor(itemView.context, tempColor))
 
-            if (meal.info.isNullOrEmpty()) {
+            if (item.meal.info.isNullOrEmpty()) {
```
KaneCheshire, line 50: I think there might be a linter warning on this line about using isNullOrEmpty when `info` isn't nullable

shared/src/commonMain/kotlin/com/kcc/kmmhackathon/shared/utility/LocationUtil.kt
```diff
@@ -20,7 +20,10 @@ class LocationUtil {
         val c: Double = 2 * atan2(sqrt(a), sqrt(1 - a))
         val distance: Double = earthRadius * c
 
-        return if (inMiles) distance.round(2) else distance.toKilometres().round(2)
+        return when (distanceUnit) {
+            0 -> distance.toKilometres().round(2)  // kilometres
+            else -> distance.round(2)  // miles
```
dsmith42, line 25: Decided to use when here so we could add more units of measure if desired.

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/view/FindMealActivity.kt
KaneCheshire: Nope doesn't look like it does!

gradle.properties
dsmith42: While Roger and I were hacking about with location we were getting some import errors / compatibility issues.  A search on google suggested adding Jetifier.  [Droid docs](https://developer.android.com/studio/command-line/jetifier)

Though I may try removing it and seeing whether it is still needed

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/adapter/MealsAdapter.kt
```diff
@@ -39,28 +39,28 @@ class MealsAdapter(var mealsList: List<Meal>) :
...
                 infoView.isVisible = false
             } else {
-                infoView.text = "Info: ${meal.info}"
+                infoView.text = "Info: ${item.meal.info}"
             }
 
             // TODO calculate distance (this could be done in the shared layer)
```
KaneCheshire: Can this todo be removed?

dsmith42: 👍

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/view/FindMealActivity.kt
```diff
@@ -1,39 +1,68 @@
...
     private lateinit var progressBarView: FrameLayout
     private lateinit var swipeRefreshLayout: SwipeRefreshLayout
 
-    private val sdk = MealsSDK()
-    private val mealsAdapter = MealsAdapter(listOf())
+    private val fusedLocationProviderClient: FusedLocationProviderClient by lazy {
+        LocationServices.getFusedLocationProviderClient(this)
+    }
+
+    private val permissionResultParser: PermissionResultParser by lazy { PermissionResultParser() }
```
KaneCheshire, line 44: Don't think this one needs to be lazy, fusedLocationProviderClient needs to be because you can't pass `this` as the context before onCreate, but since PermissionResultParser doesn't need a context or anything, it's safe to just do `val permissionResultParser = PermissionResultParser()`

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/view/FindMealActivity.kt
```diff
@@ -42,7 +71,99 @@ class FindMealActivity : AppCompatActivity() {
             swipeRefreshLayout.isRefreshing = false
             displayMeals(true)
         }
-        displayMeals(false)
+
+        checkLocationPermission()
+    }
+
+    private fun checkLocationPermission() {
```
KaneCheshire, line 78: If we're only ever requesting location after requesting permissions then we don't need to check here. I know you get linter warnings about checking permissions, but it's safe to silence them (option+return) because we know we're only ever requesting after the user has granted permissions in the onRequestPermissionsResult callback.

Also another small thing but the function is called checkLocationPermission but it also updates the last location

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/view/FindMealActivity.kt
```diff
@@ -42,7 +71,99 @@ class FindMealActivity : AppCompatActivity() {
...
+        }
+        fusedLocationProviderClient.requestLocationUpdates(locationRequest, locationCallback, null)
+    }
+
+    private fun setupLocationUpdates(locationRequest: LocationRequest) {
+        val builder = LocationSettingsRequest.Builder()
+            .addLocationRequest(locationRequest)
+
+        val client = LocationServices.getSettingsClient(this)
+        val task = client.checkLocationSettings(builder.build())
```
KaneCheshire, line 132: Hmm is this really setting up location updates, or is it just checking permissions? Since we're only requesting permissions on a successful permissions request, is it still needed?

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/adapter/MealsAdapter.kt
kirstybutler: It would be better if it was just `isEmpty()`?

PR #45, opened 2020-11-19
API update that enables user location to calculate distance

dsmith42:
Enables frontend to add user location to api call and retrieve a list with distances to meals in response

Separated this out to unblock Gareths work on API

### DIFF
--- a/shared/src/commonMain/kotlin/com/kcc/kmmhackathon/shared/utility/LocationUtil.kt
+++ b/shared/src/commonMain/kotlin/com/kcc/kmmhackathon/shared/utility/LocationUtil.kt
@@ -7,7 +7,7 @@ typealias Degree = Double
 
 class LocationUtil {
 
-    fun getDistance(lat1: Double, lon1: Double, lat2: Double, lon2: Double, inMiles: Boolean): Double {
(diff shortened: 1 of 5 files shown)


shared/src/commonMain/kotlin/com/kcc/kmmhackathon/shared/utility/LocationUtil.kt
```diff
@@ -7,7 +7,7 @@ typealias Degree = Double
 
 class LocationUtil {
 
-    fun getDistance(lat1: Double, lon1: Double, lat2: Double, lon2: Double, inMiles: Boolean): Double {
+    fun getDistance(lat1: Double, lon1: Double, lat2: Double, lon2: Double, distanceUnit: Int): Double {
```
KaneCheshire, line 10: Could DistanceUnit be used as the type instead of Int, or does that cause issues with the shared code?

KaneCheshire (approved):

kirstybutler (approved):

shared/src/commonMain/kotlin/com/kcc/kmmhackathon/shared/utility/LocationUtil.kt
dsmith42: I wasn't sure whether it would play nice on both platforms, so was keeping it agnostic in first instance.

PR #48, opened 2020-11-20
filtering meals in kmm

Gmiller290488:
I added code for Date in the shared code because all the libraries for it were causing me problems and I didn't need that much.
The filtering now happens in shared area.


iosHackthonApp/Modules/Features/MealList/Sources/MealList/MealListView.swift
```diff
@@ -14,58 +14,64 @@ public struct MealListView: View {
...
+                                dismissButton: .default(Text(Strings.Common.ok)))
+                        case .error:
+                            return Alert(
+                                title: Text(Strings.Common.sorry),
+                                message: Text(Strings.Common.ErrorAlert.message),
+                                dismissButton: .default(Text(Strings.Common.ok)))
+                        }
                     }
+                } else {
+                    Text("No meals found") // TODO: Do something nicer than this
```
Gmiller290488: Just a temporary thing so we can differentiate easier between no meals and problems with the API. We should make a nicer ui in the future for this scenario probably

iosHackthonApp/Modules/Features/MealList/Sources/MealList/MealListViewModel.swift
```diff
@@ -40,9 +40,11 @@ public final class MealListViewModel: ObservableObject {
                            completionHandler: { meals, error in
             guard
                 let meals = meals,
-                error == nil else { return }
+                error == nil else {
+                assertionFailure(String(describing: error?.localizedDescription))
```
Gmiller290488: Same as above really. Just helpful for knowing whether there's no meals or a problem

shared/src/androidMain/kotlin/com/kcc/kmmhackathon/shared/utility/DateKMM.kt
```diff
@@ -0,0 +1,5 @@
+package com.kcc.kmmhackathon.shared.utility
+
+import java.util.*
+
+actual typealias DateKMM = Date
```
Gmiller290488: This java Date class fully matches the `expect` so we don't need to add anything else 🎉

shared/src/commonMain/kotlin/com/kcc/kmmhackathon/shared/network/MealApi.kt
```diff
@@ -5,6 +5,7 @@ import com.kcc.kmmhackathon.shared.entity.MealWithDistance
 import com.kcc.kmmhackathon.shared.entity.Quantity
 import com.kcc.kmmhackathon.shared.utility.DistanceUnit
 import com.kcc.kmmhackathon.shared.utility.LocationUtil
+import com.kcc.kmmhackathon.shared.utility.DateKMM
```
Gmiller290488: Had to call it DateKMM because Xcode gets itself all confused when I called it Date - which is to be expected, feel free to suggest a better name 🤷

shared/src/commonMain/kotlin/com/kcc/kmmhackathon/shared/network/MealApi.kt
```diff
@@ -45,11 +46,16 @@ class MealApi {
...
+                userLon,
+                meal.locationLat.toDouble(),
+                meal.locationLong.toDouble(),
+                distanceUnit
             )
         }
-        // TODO: Filter meals that have expired here
-        return meals.sortedWith(compareBy { it.distance })
+        return meals
+            .filterNot { DateKMM( it.expiryDate.toLong()).compareTo(DateKMM()) < 0 }
```
Gmiller290488: If the expiryDate is before the current date, -1 is returned

shared/src/commonMain/kotlin/com/kcc/kmmhackathon/shared/utility/DateKMM.kt
```diff
@@ -0,0 +1,18 @@
...
+    constructor(time: Long)
+
+    /**
+     * Compare this instance with the other instance.
+     *
+     * @param other: Another date instance.
+     * @return 0 if the other is equal to this instance; a value less than 0 if this instance is before the other;
+     *          and a value greater than 0 if this instance is after the other.
+     */
+    fun compareTo(other: DateKMM): Int
```
Gmiller290488: This is the documentation for the Java data compareTo function.
It's not the easiest function to understand, we could add our own compare function that returns a bool but I think then it would take a custom implementation of Date on Android side - but happy to do it if everyone thinks this is a bit gross

raynelle03 (approved):

shared/src/iosMain/kotlin/com/kcc/kmmhackathon/shared/utility/DateKMM.kt
```diff
@@ -0,0 +1,32 @@
...
+
+    actual constructor(time: Long) {
+        date = NSDate.dateWithTimeIntervalSince1970(time.toDouble())
+    }
+
+    constructor(platformDate: NSDate) {
+        date = platformDate
+    }
+
+    actual fun compareTo(other: DateKMM): Int {
```
aabolade: Is there any way we can add tests for this logic?

Gmiller290488: Yeah currently doing it 👍

Gmiller290488: Done

iosHackthonApp/Modules/Features/MealList/Sources/MealList/MealListView.swift
```diff
@@ -14,58 +14,64 @@ public struct MealListView: View {
...
+                .padding()
+            }
             
             if viewModel.showMap == 0 {
-                ScrollView {
-                    LazyVStack {
-                        ForEach((0 ..< viewModel.meals.count), id: \.self) {
-                            MealRow(listViewModel: viewModel, rowViewModel: MealRowViewModel(meal: viewModel.meals[$0]))
-                                .padding()
+                if viewModel.meals.count > 0 {
```
KaneCheshire: How about
```swift
if viewModels.meals.isEmpty {
  Text("No meals found")
} else {
 ... etc
}
```

iosHackthonApp/Modules/Features/MealList/Sources/MealList/MealListViewModel.swift
KaneCheshire: I think it's okay to force unwrap the error here to make this more readable since you've already checked that it's not nil:

`assertionFailure(error!.localizedDescription)` or `assertionFailure(error?.localizedDescription ?? "")`

shared/src/androidMain/kotlin/com/kcc/kmmhackathon/shared/utility/DateKMM.kt
KaneCheshire: Nice!

shared/src/commonMain/kotlin/com/kcc/kmmhackathon/shared/network/MealApi.kt
KaneCheshire: You can disambiguate by including the module name before when using Date like `Foundation.Date` or `Swift.Date` (can't remember which) and `shared.Date` but that might get tedious, so makes sense to typealias. Not super keen on the name, but it works. Maybe something like SharedDate, XDate (as in, cross platform date)

shared/src/commonMain/kotlin/com/kcc/kmmhackathon/shared/utility/DateKMM.kt
KaneCheshire: I think there are some comparison APIs on the iOS side that return an enum (basically an Int) so I think it's okay to leave, might be nice to wrap up into something a bit more Swifty but fine as-is too :)

shared/src/commonMain/kotlin/com/kcc/kmmhackathon/shared/network/MealApi.kt
Gmiller290488: I might go with sharedDate 👍

iosHackthonApp/Modules/Features/MealList/Sources/MealList/MealListViewModel.swift
```diff
@@ -40,9 +40,11 @@ public final class MealListViewModel: ObservableObject {
                            completionHandler: { meals, error in
             guard
                 let meals = meals,
-                error == nil else { return }
+                error == nil else {
+                assertionFailure(String(describing: error!.localizedDescription))
```
KaneCheshire: You don't need the `String(describing` bit now since that was just to fix the optional :)

dsmith42:
Looks good, really great to be lifting more of the logic up into the shared layer 🚀

dsmith42 (approved):
Looks good, really great to be lifting more logic into the shared layer.  Sharing is caring 🥳

iosHackthonApp/Modules/Features/MealList/Sources/MealList/MealListViewModel.swift
Gmiller290488: Dopey of me 🤦

shared/src/commonMain/kotlin/com/kcc/kmmhackathon/shared/utility/extensions/SharedDate+Extensions.kt
```diff
@@ -0,0 +1,15 @@
...
+    return this.compareTo(other) > 0
+}
+
+fun SharedDate.isBefore(other: SharedDate): Boolean {
+    return this.compareTo(other) < 0
+}
+
+fun SharedDate.isEqual(other: SharedDate): Boolean {
+    return this.compareTo(other) == 0
+}
```
Gmiller290488, line 15: Added these because every time I looked at the `compareTo` code in the API file it made me go 🤯  and I was the one who wrote it

shared/src/commonMain/kotlin/com/kcc/kmmhackathon/shared/network/MealApi.kt
```diff
@@ -45,11 +44,16 @@ class MealApi {
...
+                userLon,
+                meal.locationLat.toDouble(),
+                meal.locationLong.toDouble(),
+                distanceUnit
             )
         }
-        // TODO: Filter meals that have expired here
-        return meals.sortedWith(compareBy { it.distance })
+        return meals
+            .filter { SharedDate().isBefore(SharedDate(it.expiryDate.toLong())) }
```
Gmiller290488, line 45: filter meals where the date now, is before the expiry date.

I think this is easier to understand now

shared/src/commonMain/kotlin/com/kcc/kmmhackathon/shared/utility/DateKMM.kt
Gmiller290488: I extended ShareDate to have `isBefore`, `isAfter` and `isEqual` functions to make this a bit easier

PR #51, opened 2020-11-23
Feature/authentication

aabolade:
This PR deals with adding objects responsible for making calls to the Firebase Authentication REST API. The [project](https://console.firebase.google.com/u/1/project/communitychef-61133/overview) can be found here. Let me know if you can't access it.


shared/src/commonMain/kotlin/com/kcc/kmmhackathon/shared/entity/Entity.kt
```diff
@@ -38,4 +38,16 @@ data class MealWithDistance(
     val meal: Meal,
     @SerialName("distance")
     val distance: Double
+)
+
+@Serializable
+data class FirebaseAuthenticationResponse(
```
KaneCheshire, line 37: Do the properties on this data class need to be nullable? Just looking at the ones above that aren't. Other than that, this PR looks nice to me! Nice and tidy

KaneCheshire (approved):

KaneCheshire (approved):

raynelle03:
Looks like Tests are failing.

KaneCheshire:
Ah yes, it's failing to compile because it's expecting some values that aren't nullable now

Gmiller290488 (approved):

PR #53, opened 2020-11-23
Fix/ds android dates

dsmith42:
Adds some date handling functionality to the Android app.

Also, updates the json-server db file as they were all self filtering out to give 2 meals that will expire in the future (Dec 13th)


shared/src/androidMain/kotlin/com/kcc/kmmhackathon/shared/utility/SharedDate.kt
```diff
@@ -1,5 +1,15 @@
...
+import java.time.ZoneId
+import java.time.format.DateTimeFormatter
 import java.util.*
 
-actual typealias SharedDate = Date
\ No newline at end of file
+actual typealias SharedDate = Date
+
+fun timeInMillisToReadableString(timeInMillis: Long): String {
+    val formatter = DateTimeFormatter.ofPattern("EEE dd MMM YYYY HH:mm")
```
dsmith42, line 12: Found that EEE is the way to get the shorter DayOfWeek in formatter.

shared/src/db.json
```diff
@@ -5,7 +5,7 @@
       "name": "Tomato pasta",
       "quantity": 1,
       "availableFromDate": "1605802429",
-      "expiryDate": "1405802429",
+      "expiryDate": "1607849537537",
       "info": "Vegetarian",
```
dsmith42, line 9: These timestamps were in the 70s so was the whole list was filtered out.  Have updated 2 of them so that the filter system will still have something to run on.

Gmiller290488 (commented):
We've been up to the same thing 😬

shared/src/db.json
Gmiller290488: Depends 🤔 if you are converting using seconds or milliseconds.
For sure it was in the past but that expiry date is in 2014 if using seconds.

We will need to make sure we align how we are creating these timestamps from the addMeal screen so we don't end up with botched dates on one platform or another.  Let's catch up tomorrow morning

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/adapter/MealsAdapter.kt
```diff
@@ -11,7 +11,12 @@ import androidx.core.view.isVisible
 import androidx.recyclerview.widget.RecyclerView
 import com.kcc.kmmhackathon.shared.entity.Meal
 import com.kcc.kmmhackathon.shared.utility.DistanceUnit
+import com.kcc.kmmhackathon.shared.utility.timeInMillisToReadableString
 import com.kinandcarta.lib.find.meal.R
+import java.time.Instant
+import java.time.LocalDateTime
+import java.time.ZoneId
+import java.time.format.DateTimeFormatter
```
Gmiller290488, line 19: Are these used?

shared/src/androidMain/kotlin/com/kcc/kmmhackathon/shared/utility/SharedDate.kt
roger-tan: I haven't tested but it will be great if we could use `dateStyle` and `timeStyle` with maybe with `DateTimeFormatter.ofLocalizedDateTime`

PR #55, opened 2020-11-24
Adding Hilt and refactoring to ViewModel

KaneCheshire:
This PR sets the foundations for using Hilt and view models in the app. I've refactored the FindMealFragment to make the fragment responsible for handling interactions and reacting to the current view model state, and the view model is responsible for grabbing the user's location and fetching meals from the API.

The way permissions work on Android requires you to listen to a callback in the activity or fragment, so what I've done is make it so that we request permissions and only start the VM listening to locations once we get a successful callback. If the user has already granted permissions, then the callback is invoked immediately with success (like on iOS).


.idea/misc.xml
```diff
@@ -1,5 +1,10 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
+  <component name="EntryPointsManager">
```
KaneCheshire, line 3: Autogenerated for me, although I think Voicu said to ignore .idea files. Fine for now I guess

androidHackathonApp/build.gradle.kts
```diff
@@ -2,7 +2,8 @@ plugins {
     id("com.android.application")
     kotlin("android")
     id("kotlin-android-extensions")
```
KaneCheshire, line 4: This can probably be `kotlin("android-extensions")` but I'll leave for now

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/injection/FindMealModule.kt
```diff
@@ -0,0 +1,20 @@
...
+import dagger.hilt.android.components.FragmentComponent
+import dagger.hilt.android.qualifiers.ActivityContext
+
+@Module
+@InstallIn(FragmentComponent::class)
+class FindMealModule {
+
+    companion object {
+        @Provides
+        fun provideFusedLocationClient(@ActivityContext context: Context) =
```
KaneCheshire, line 17: Here's where some of the magic happens. Hilt uses this @Provides function to create an instance of FusedLocationProvider whenever somewhere that injects it needs it (like in the view model constructor). By adding @ActivityContext into this function, it injects the activity context (not app context) into the function for me so I can use it to create the provider. You can also use @ApplicationContext here

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/view/FindMealFragment.kt
```diff
@@ -1,68 +1,37 @@
...
-import com.kcc.kmmhackathon.shared.utility.DistanceUnit
+import com.kinandcarta.lib.find.meal.extension.requestFineLocationPermission
+import com.kinandcarta.lib.find.meal.extension.showToast
 import com.kinandcarta.lib.find.meal.utility.PermissionResultParser
-import kotlinx.coroutines.cancel
+import com.kinandcarta.lib.find.meal.viewmodel.FindMealViewModel
+import dagger.hilt.android.AndroidEntryPoint
 
+@AndroidEntryPoint
 class FindMealFragment : Fragment() {
```
KaneCheshire, line 25: I've moved quite a lot out of this file and into the VM, the adapter and extensions to try and keep it as neat as possible. The result is that the view (the fragment) is mostly just responsible for handling interactions and reacting to different states that the view model is in

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/view/FindMealFragment.kt
```diff
@@ -1,68 +1,37 @@
...
-        override fun onLocationResult(locationResult: LocationResult?) {
-            lastLocation = locationResult?.lastLocation ?: return
-        }
-    }
-
-    private lateinit var viewModel: FindMealViewModel
+    private val progressBarView: FrameLayout by lazy { requireView().findViewById(R.id.progressBar) }
+    private val swipeRefreshLayout: SwipeRefreshLayout by lazy { requireView().findViewById(R.id.swipeContainer) }
+    private val viewModel: FindMealViewModel by viewModels()
+    private val mealsAdapter = MealsAdapter()
```
KaneCheshire, line 29: Down from 14 properties to 4 :)

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/view/FindMealFragment.kt
```diff
@@ -73,141 +42,61 @@ class FindMealFragment : Fragment() {
...
-                Manifest.permission.ACCESS_FINE_LOCATION
-            ) !== PackageManager.PERMISSION_GRANTED
-        ) {
-            requestFineLocationPermission(LOCATION_PERMISSION_REQUEST_CODE)
-        }
-    }
-
-    fun requestFineLocationPermission(requestCode: Int = 0) {
-        requestPermissions(arrayOf(Manifest.permission.ACCESS_FINE_LOCATION), requestCode)
+        setupUIWithView(view)
```
KaneCheshire: Don't actually need to pass in the view here, can just do `requireView()` in the function, will update

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/view/FindMealFragment.kt
```diff
@@ -73,141 +42,61 @@ class FindMealFragment : Fragment() {
...
+        viewModel.state.observe(viewLifecycleOwner, ::onStateChanged)
     }
 
+    @SuppressLint("MissingPermission")
     override fun onRequestPermissionsResult(
         requestCode: Int,
         permissions: Array<out String>,
         grantResults: IntArray
     ) {
+        val permissionResultParser = PermissionResultParser()
```
KaneCheshire, line 54: Moved this to here since we don't need the parser anywhere else in the file, so don't really need to clutter up the properties by adding it as a property

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/view/FindMealFragment.kt
```diff
@@ -73,141 +42,61 @@ class FindMealFragment : Fragment() {
...
         if (permissionResultParser.isFineLocationPermissionsGranted(permissions, grantResults)) {
-            getLastLocation()
+            viewModel.startUpdatingLocation()
```
KaneCheshire, line 56: the `startUpdatingLocation` function requires location permissions, but since we're doing this in this callback function we can just suppress the linting since we're confident that they've given us permissions

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/view/FindMealFragment.kt
```diff
@@ -73,141 +42,61 @@ class FindMealFragment : Fragment() {
...
         } else {
-            // TODO Ask user to update settings / location settings are required
```
KaneCheshire, line 118: It shows a toast here now (can be hard to see in the diff unles you use a side-by-side)

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/view/FindMealFragment.kt
```diff
@@ -73,141 +42,61 @@ class FindMealFragment : Fragment() {
...
-            progressBarView.isVisible = false
         }
     }
 
     companion object {
-        private val LOCATION_PERMISSION_REQUEST_CODE = 1
-        private val REQUEST_CHECK_SETTINGS = 2
-        fun newInstance() = FindMealFragment()
+        private const val LOCATION_PERMISSION_REQUEST_CODE = 1
+        private const val REQUEST_CHECK_SETTINGS = 2 // TODO: Need to re-add this check somewhere?
```
KaneCheshire: This isn't being used because I took the code out, wasn't sure if we definitely need it still, but we can re-add if necessary. If not, I'll remove this, I just left it here to remind us to think about it

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/viewmodel/FindMealViewModel.kt
```diff
@@ -1,7 +1,86 @@
...
+import com.kcc.kmmhackathon.shared.MealsSDK
+import com.kcc.kmmhackathon.shared.entity.Meal
+import com.kcc.kmmhackathon.shared.utility.DistanceUnit
+import kotlinx.coroutines.launch
 
-class FindMealViewModel : ViewModel() {
-    // TODO: Implement the ViewModel
+typealias Meals = List<Meal>
+
+class FindMealViewModel @ViewModelInject constructor(
```
KaneCheshire, line 21: `@ViewModelInject` is a special version of `@Inject` which we'll see more as we start to use Hilt more. Without `ViewModelInject` we need to jump through some hoops to inject directly into the constructor, so Dagger is more annoying, but thankfully they added this in Hilt so we can just do `by viewModels()` in the Fragment

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/viewmodel/FindMealViewModel.kt
```diff
@@ -1,7 +1,86 @@
...
+    private val fusedLocationClient: FusedLocationProviderClient
+) : ViewModel() {
+
+    sealed class State {
+        object LoadingMeals : State()
```
KaneCheshire, line 26: We could add a "refreshing meals" state to this if we needed to differentiate in the UI between initial load and pull to refresh etc

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/viewmodel/FindMealViewModel.kt
```diff
@@ -1,7 +1,86 @@
...
+            locationCallback,
+            null
+        )
+    }
+
+    fun updateMeals(needReload: Boolean) {
+        val location = lastLocation ?: return
+        val distanceUnit = DistanceUnit.miles
+        _state.value = State.LoadingMeals
+        viewModelScope.launch {
```
KaneCheshire, line 68: Android VMs come with a scope that automatically gets cancelled on clear I believe

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/view/FindMealFragment.kt
KaneCheshire: 3!

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/injection/FindMealModule.kt
dsmith42: Niiice, that will set us up for reuse in Maps Fragment too

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/view/FindMealFragment.kt
dsmith42: I think with the refactoring that we can remove it now.

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/viewmodel/FindMealViewModel.kt
dsmith42: That could be a nice touch.  I think we should also consider some UI for when there are no meals to list

dsmith42: Use of State here is really cool 🤩

lib-theming/build.gradle.kts
```diff
@@ -25,6 +27,13 @@ dependencies {
     implementation("androidx.recyclerview:recyclerview:$recyclerVersion")
     implementation("androidx.swiperefreshlayout:swiperefreshlayout:1.1.0")
     implementation("androidx.cardview:cardview:1.0.0")
+
+    // Hilt
+    implementation("com.google.dagger:hilt-android:2.28-alpha")
```
dsmith42, line 32: Should we lift version out as it is used in multiple places?  (Not sure of preferred Droid conventions)

dsmith42 (approved):
Great job Kane, so much cleaner!

Gmiller290488 (approved):

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/viewmodel/FindMealViewModel.kt
KaneCheshire: That's a good shout! It's easy to add an empty state now, just need to add a new object and then handle it in the UI

lib-theming/build.gradle.kts
KaneCheshire: yesss I really want to, but I'm trying to see if there's a way to only have to define the entire dependency once rather than in each module because it's already getting tedious

roger-tan (approved):

PR #57, opened 2020-11-24
Remove forced reload

Gmiller290488:
I saw this as a TODO in someones PR and it was something I kept meaning to get rid of but kept forgetting about...

In case you were wondering the tutorial we used as an outline for the project made use of it and it's been in our project since day 1 because of that 😅


raynelle03 (approved):

KaneCheshire (approved):

dsmith42 (approved):
Nice one 😃

PR #58, opened 2020-11-24
Add meal button in nav bar

Gmiller290488:
Added a plus button in the meal list and map view which opens the add Meal screen as a sheet.

I went with displaying the AddMeal in a sheet because it felt weird to push it... it would need a back arrow and you could change tabs but still be on the AddMeal screen when you switched tabs back again and I think that this plus button here is only a temporary situation anyway :-)

Also created the package for the Settings, refactored some duplicated code, and removed some unused code

small-addmeal


iosHackthonApp/Modules/Features/MealList/Sources/MealList/MealListView.swift
```diff
@@ -26,13 +27,22 @@ public struct MealListView: View {
...
+                                    }) {
+                                        Image(systemName: Strings.Common.Images.add)
+                                    },
+                                trailing:
+                                    Button(action: {
+                                        self.viewModel.loadMeals()
+                                    }, label: {
+                                        Image(systemName: Strings.MealListScreen.Images.reload)
+                                    }))
+                .foregroundColor(.white)
```
Gmiller290488: will fix this indentation - just want to see if CI passes 😬

iosHackthonApp/Modules/Features/MealList/Sources/MealList/MealListViewModel.swift
```diff
@@ -40,7 +41,7 @@ public final class MealListViewModel: ObservableObject {
             guard
                 let meals = meals,
                 error == nil else {
-                assertionFailure(error!.localizedDescription)
+                print(error!.localizedDescription)
```
KaneCheshire, line 46: Hmm does this mean we should maybe be showing some sort of error state to the user?

KaneCheshire (approved):

iosHackthonApp/Modules/Features/MealList/Sources/MealList/MealListViewModel.swift
Gmiller290488: Yeah, we should show an alert or something. I added this just as a catch to remind me that I didn't start the `json-server` yet 😬 but I think that assertion was causing my CI problems last night/this morning which is why I changed it to a print statement

iosHackthonApp/Modules/Features/AddMeal/Sources/AddMeal/AddMealView.swift
```diff
@@ -46,46 +46,48 @@ public struct AddMealView: View {
...
-                    AddressTextFieldGroup(locationManager: locationManager, address: $address, latitude: $latitude, longitude: $longitude)
-                    Spacer()
-                    GeometryReader { geometry in
-                        Button(action: {
-                            postMeal(meal: createMeal())
-                        }) {
-                            Text(Strings.AddMealScreen.addMeal)
-                                .modifier(AddButtonStyle(width: geometry.size.width,
-                                                         backgroundColor: buttonColor))
+        NavigationView {
```
Gmiller290488, line 49: The only change in here is that I added the NavigationView

raynelle03 (approved):

PR #60, opened 2020-11-25
Feature/ds droid add reserve

dsmith42:
Lifts the reserve logic up to KMM
Wires up the Android reserve button

### DIFF
--- a/lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/adapter/MealsAdapter.kt
+++ b/lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/adapter/MealsAdapter.kt
@@ -16,7 +16,8 @@ import com.kinandcarta.lib.find.meal.viewmodel.Meals
 
 class MealsAdapter(
     private var mealsList: List<Meal> = listOf(),
-    private var distanceUnit: DistanceUnit = DistanceUnit.miles
+    private var distanceUnit: DistanceUnit = DistanceUnit.miles,
+    private var clickListener: (String) -> Unit
 ) :
     RecyclerView.Adapter<MealsAdapter.MealsViewHolder>() {
 
@@ -82,10 +83,7 @@ class MealsAdapter(
             )
             if (hasPortions) {
                 reserveButton.setOnClickListener {
-                    // TODO link up to sdk and adjust quantity and show user reservation code
-                    val id = meal.id
-                    val code = id.subSequence(id.length - 4, id.length)
-                    Log.i("Reserve button tapped", "${meal.name} reservation code ${code}")
+                    clickListener(meal.id)
                 }
             }
         }
--- a/lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/adapter/MealsAdapter.kt
+++ b/lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/adapter/MealsAdapter.kt
@@ -16,7 +16,7 @@ import com.kinandcarta.lib.find.meal.viewmodel.Meals
 class MealsAdapter(
     private var mealsList: List<Meal> = listOf(),
     private var distanceUnit: DistanceUnit = DistanceUnit.miles,
-    private var clickListener: (String) -> Unit
+    private val clickListener: (String, Int) -> Unit
 ) :
     RecyclerView.Adapter<MealsAdapter.MealsViewHolder>() {
 
@@ -30,7 +30,7 @@ class MealsAdapter(
     }
 
     override fun onBindViewHolder(holder: MealsViewHolder, position: Int) {
-        holder.bindData(mealsList[position])
+        holder.bindData(mealsList[position], position)
     }
 
     fun submit(meals: Meals, unit: DistanceUnit) {
@@ -49,7 +49,7 @@ class MealsAdapter(
         private val portionsView = itemView.findViewById<TextView>(R.id.mealPortions)
         private val reserveButton = itemView.findViewById<Button>(R.id.reserveButton)
 
-        fun bindData(meal: Meal) {
+        fun bindData(meal: Meal, position: Int) {
             nameView.text = meal.name
             val tempString = if (meal.hot) "Hot" else "Cold"
             val tempColor = if (meal.hot) R.color.colorHot else R.color.colorCold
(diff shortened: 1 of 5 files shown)


lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/view/FindMealFragment.kt
```diff
@@ -26,7 +26,8 @@ class FindMealFragment : Fragment() {
 
     private val progressBarView: FrameLayout by lazy { requireView().findViewById(R.id.progressBar) }
     private val viewModel: FindMealViewModel by viewModels()
-    private val mealsAdapter = MealsAdapter()
+
+    private val mealsAdapter = MealsAdapter() { id -> viewModel.reserveAMeal(id)}
```
dsmith42: Adding the click listener callBack

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/viewmodel/FindMealViewModel.kt
```diff
@@ -83,4 +85,17 @@ class FindMealViewModel @ViewModelInject constructor(
...
+
+    fun reserveAMeal(id: String) {
+        viewModelScope.launch {
+            kotlin.runCatching {
+                sdk.reserveAMeal(id)
+            }.onSuccess {
+                Log.i("Reserve a meal", "${it.quantity}")
+                updateMeals()
+            }.onFailure {
+                _state.value = State.Failed(Failure.ReserveAMealFailed(it))
```
dsmith42, line 102: Making use of Kane's failed state addition 🙏

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/view/FindMealFragment.kt
```diff
@@ -86,12 +89,21 @@ class FindMealFragment : Fragment() {
         mealsAdapter.submit(state.meals, state.distanceUnit)
     }
 
+    private fun onReservedMeal(state: FindMealViewModel.State.ReservedMeal) {
+        val id = state.meal.id
+        val code = id.subSequence(id.length - 4, id.length)
```
KaneCheshire: hmm should this logic be in the vm?

shared/src/commonMain/kotlin/com/kcc/kmmhackathon/shared/network/MealApi.kt
```diff
@@ -62,6 +62,14 @@ class MealApi {
...
     }
 
+    suspend fun reserveAMeal(id: String): Meal {
+        var meal = getMeal(id)
+        if (meal.quantity > 0) {
+            return patchMeal(id, meal.quantity - 1)
+        }
+        return meal
+    }
+
```
Gmiller290488: If for example a user tries to reserve a meal that is available on their device but in actuality the quantity is at zero (because they didn't refresh the list), then we are just returning the meal back. 
So _I think_ then a toast would be displayed with a collection code?  Because it's technically success state.

We could make the return optional and return null if we don't patch and then catch that?
Or maybe this functionality would be better off being in the MealsSDK (and same for the function that sorts and filters all the meals too 🤔 ) and we could throw an error from there? 
We can't do the same as what I did on ios because when the meal is returned with 0 for quantity, we can't distinguish between it being zero because before it was 1, or it's zero because it was zero all along and we didn't make the PATCH.

Anyway I'm just rambling and putting my thoughts down

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/viewmodel/FindMealViewModel.kt
```diff
@@ -26,10 +27,12 @@ class FindMealViewModel @ViewModelInject constructor(
         object LoadingMeals : State()
         data class LoadedMeals(val meals: Meals, val distanceUnit: DistanceUnit) : State()
         data class Failed(val failure: Failure) : State()
+        data class ReservedMeal(val meal: Meal) : State()
```
KaneCheshire: Very very minor but i tend to keep failure as the last case/state, and then keep the other states as close to the order they'll be encountered (i.e loading, loaded, reserved, failed)

shared/src/commonMain/kotlin/com/kcc/kmmhackathon/shared/network/MealApi.kt
```diff
@@ -62,6 +62,14 @@ class MealApi {
         }
     }
 
+    suspend fun reserveAMeal(id: String): Meal {
+        var meal = getMeal(id)
+        if (meal.quantity > 0) {
```
KaneCheshire: i think there should be a kotlin equivalent to isEmpty in Swift, i think they even have isNotEmpty()

shared/src/commonMain/kotlin/com/kcc/kmmhackathon/shared/network/MealApi.kt
```diff
@@ -62,6 +62,14 @@ class MealApi {
         }
     }
 
+    suspend fun reserveAMeal(id: String): Meal {
+        var meal = getMeal(id)
```
KaneCheshire: Does this need to be var? i can't see it being reassigned anywhere in the function

lib-find-meal/src/main/java/com/kinandcarta/lib/find/meal/adapter/MealsAdapter.kt
```diff
@@ -16,7 +15,8 @@ import com.kinandcarta.lib.find.meal.viewmodel.Meals
 
 class MealsAdapter(
     private var mealsList: List<Meal> = listOf(),
-    private var distanceUnit: DistanceUnit = DistanceUnit.miles
+    private var distanceUnit: DistanceUnit = DistanceUnit.miles,
+    private var clickListener: (String) -> Unit
```
KaneCheshire: Think this can be val since it's not going to be reassigned? Although I might be wrong

KaneCheshire (approved):

Gmiller290488 (approved):

roger-tan (approved):

PR #63, opened 2020-11-26
Add listView to bottom of map

Gmiller290488:
Happy to change any aspect of this so tell me your thoughts!
I had wanted to implement that dragging when at the top of the scrollview would dismiss the view but in all honesty the way I had managed to get it working was hack-y and a bit gross.  I might revisit it in the future 🤔 

Hnet-image


iosHackthonApp/Modules/Shared/Components/Sources/Components/BottomSheet.swift
```diff
@@ -0,0 +1,81 @@
...
+                        }
+                } else {
+                    Button(action: {
+                        isOpen.toggle()
+                    }, label: {
+                        Text(labelText)
+                            .foregroundColor(ColorManager.appPrimary)
+                    })
+                    .padding()
+                }
```
Gmiller290488: Will create these views elsewhere to make this a bit easier to read

iosHackthonApp/Modules/Shared/Components/Sources/Components/BottomSheet.swift
```diff
@@ -0,0 +1,89 @@
...
+                content
+            }
+            .frame(width: geometry.size.width, height: maxHeight, alignment: .top)
+            .background(Color.white)
+            .cornerRadius(Constants.cornerRadius)
+            .frame(height: geometry.size.height, alignment: .bottom)
+            .offset(y: max(offset + translation, 0))
+            .animation(.easeInOut)
+            .gesture(
+                DragGesture().updating($translation) { value, state, _ in
```
KaneCheshire: Could break this out into a private property like `showListButton`? I like those because they make it much easier to read

```swift
private var drag: some Gesture {
  DragGesture().updating ...etc
 }
```

iosHackthonApp/Modules/Shared/Components/Sources/Components/BottomSheet.swift
```diff
@@ -0,0 +1,89 @@
...
+import Theming
+
+fileprivate enum Constants {
+    static let cornerRadius: CGFloat = 16
+    static let indicatorWidth: CGFloat = 60
+    static let indicatorHeight: CGFloat = 6
+    static let minHeightRatio: CGFloat = 0.3
+}
+
+public struct BottomSheetView<Content: View>: View {
```
KaneCheshire, line 19: 😍 loveeee this

KaneCheshire (approved):

iosHackthonApp/Modules/Shared/Components/Sources/Components/BottomSheet.swift
```diff
@@ -0,0 +1,93 @@
...
+    private let maxHeight: CGFloat
+    private let minHeight: CGFloat
+    private let content: Content
+    private let labelText: String
+
+    @GestureState private var translation: CGFloat = 0
+    @Binding private var isOpen: Bool
+
+    public init(isOpen: Binding<Bool>, maxHeight: CGFloat, labelText: String, @ViewBuilder content: () -> Content) {
+        self.minHeight = min(maxHeight * Constants.minHeightRatio, 50)
```
Gmiller290488, line 30: Been a bit annoying finding a value that works for all devices here tbh.
This now works though, if the maxHeight * 0.15 > 50 (which it is on the larger devices) then we use 50 for the minHeight.
This stops the first item in the list being visible on the bottom sheet before it is opened.

raynelle03 (approved):

PR #65, opened 2020-11-27
Fix/ds missing distance reserve meal

dsmith42:
Found a couple of bugs when working on the android app.

When we reserve a meal there was no distance calculation being carried out so the meal returned ended up having null for distance.

~~Changes the recyclerAdapter to use a mutable list so we can update one meal item at a time (e.g. on reserving).~~

Makes the recyclerAdapter use ListAdapter and DiffUtil to update only the items that have changed.  Useful here for when we want to reserve a meal.


feature-find-meal/src/main/java/com/kinandcarta/feature/find/meal/adapter/MealsAdapter.kt
```diff
@@ -14,7 +14,7 @@ import com.kinandcarta.feature.find.meal.R
 import com.kinandcarta.feature.find.meal.viewmodel.Meals
 
 class MealsAdapter(
-    private var mealsList: List<Meal> = listOf(),
+    private var mealsList: MutableList<Meal> = mutableListOf(),
```
dsmith42: Made it mutable so that we can update the item that changed on the reserve only, rather than having to refresh the whole list.  Which is a worse UX.

feature-find-meal/src/main/java/com/kinandcarta/feature/find/meal/view/FindMealFragment.kt
```diff
@@ -92,13 +92,14 @@ class FindMealFragment : Fragment() {
     }
 
     private fun onReservedMeal(state: FindMealViewModel.State.ReservedMeal) {
-        mealsAdapter.notifyItemChanged(state.position)
```
dsmith42, line 95: Realised this wasn't working as I had anticipated.  Due to no update actually being carried out on the list.

feature-find-meal/src/main/java/com/kinandcarta/feature/find/meal/viewmodel/FindMealViewModel.kt
```diff
@@ -46,6 +46,7 @@ class FindMealViewModel @ViewModelInject constructor(
             updateMeals()
         }
     }
+    private val distanceUnit = DistanceUnit.miles
```
dsmith42: Lifted this up as we use it in 2 places

shared/src/commonMain/kotlin/com/kcc/kmmhackathon/shared/MealsSDK.kt
```diff
@@ -45,15 +45,24 @@ class MealsSDK {
...
+            val updatedMeal = api.patchMeal(meal.id, meal.quantity - 1)
+            updatedMeal.distance = locationUtil.getDistance(
+                userLat,
+                userLon,
+                updatedMeal.locationLat.toDouble(),
+                updatedMeal.locationLong.toDouble(),
+                distanceUnit
+            )
+            return updatedMeal
         }
```
dsmith42: Adds the distance calculation to the ReserveMeal in SDK, which is needed here to avoid retuning the Meal with a null distance value.

feature-find-meal/src/main/java/com/kinandcarta/feature/find/meal/adapter/MealsAdapter.kt
KaneCheshire: Nice. I think we could also use [ListAdapter](https://developer.android.com/reference/androidx/recyclerview/widget/ListAdapter) to handle a lot of this for us but this works great as well

feature-find-meal/src/main/java/com/kinandcarta/feature/find/meal/view/FindMealFragment.kt
KaneCheshire: I think what you've changed it to is cleaner anyway, it would just be nice not to have to worry about the position at all and maybe let the adapter figure that out (which I think ListAdapter does). Not a problem for now though, since this works

KaneCheshire (approved):

feature-find-meal/src/main/java/com/kinandcarta/feature/find/meal/adapter/MealsAdapter.kt
Gmiller290488: Playing the other side here:
I'm a user that tried to reserve a meal - turns out my list is not up to date and that meal is gone so I get an alert (or toast) saying "Sorry that meal is unavailable"
and that item is updated to say as much.
I go to the next item nearby and tap to reserve, but this item is also not up to date.
I get the same alert and that one item updates to show it's unavailable.

Could be annoying?

Gmiller290488: Another thing maybe to consider is that if the user is moving.
One items distance gets updated but the rest of the list doesn't

feature-find-meal/src/main/java/com/kinandcarta/feature/find/meal/adapter/MealsAdapter.kt
```diff
@@ -39,6 +39,11 @@ class MealsAdapter(
         notifyDataSetChanged()
     }
 
+    fun updateMeal(position: Int, meal: Meal) {
+        mealsList[position] = meal
+        notifyItemChanged(position)
```
Gmiller290488: Could you not update just the quantity here, as that is what has changed, and then you wouldn't need the distance code for the reserve call?

feature-find-meal/src/main/java/com/kinandcarta/feature/find/meal/adapter/MealsAdapter.kt
dsmith42: Perhaps we need to consider how up to date the list is going to need to be to cover those scenarios?

feature-find-meal/src/main/java/com/kinandcarta/feature/find/meal/adapter/MealsAdapter.kt
dsmith42: That would do the trick and mean nothing else needs to change on API side.  Well spotted!

feature-find-meal/src/main/java/com/kinandcarta/feature/find/meal/adapter/MealsAdapter.kt
Gmiller290488: Not sure what you mean

dsmith42: What I meant was, once we have a list, after a little bit of time it may be out of date, regarding the reservation status and distance.   So perhaps we might want to think about adding a refresh list reminder or timed / event driven refresh

Gmiller290488: Yeah something to think on

feature-find-meal/src/main/java/com/kinandcarta/feature/find/meal/view/FindMealFragment.kt
dsmith42: Have converted over to use the ListAdapter 👍

feature-find-meal/src/main/java/com/kinandcarta/feature/find/meal/adapter/MealsAdapter.kt
dsmith42: Have updated to use ListAdapter diffing, allowing the List to be updated as a whole.

Gmiller290488 (approved):

PR #66, opened 2020-11-27
Updated android.yml to try and install onto a running emulator

KaneCheshire:
This will hopefully catch naughty breakages like the one I introduced yesterday, assuming that if Gradle syncs and the project compiles then all was fine 🤦

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -10,7 +10,7 @@ on:
 jobs:
   test:
 
-    runs-on: ubuntu-latest
+    runs-on: macos-latest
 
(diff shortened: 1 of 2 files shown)


KaneCheshire:
That last commit should have failed the build

Gmiller290488 (approved):

dsmith42 (approved):
Looks good.  Interesting that we need to run on MacOS

KaneCheshire:
Something about Ubuntu makes the emulator install take aaaages. They recommend using macOS to speed it up

PR #68, opened 2020-11-30
Feature/[AND] ds maps

dsmith42:
Adds maps to the android project and places markers for each meal in the meals list.

To see the maps on the device / emulator you will need to add api key to the google_maps_api.xml file.  
Roger has set one up but you'll need to add SHA-1 fingerprint to it.
https://developers.google.com/maps/documentation/android-sdk/get-api-key

Screenshot_1606730693


androidHackathonApp/src/main/java/com/kcc/kmmhackathon/androidHackathonApp/view/MapsFragment.kt
```diff
@@ -36,8 +45,67 @@ class MapsFragment : Fragment() {
...
+
+    private fun setupUI() {
         val mapFragment = childFragmentManager.findFragmentById(R.id.map) as SupportMapFragment?
         mapFragment?.getMapAsync(callback)
     }
+
+    private fun onStateChanged(state: MapsViewModel.State) {
+        when (state) {
+            MapsViewModel.State.LoadingMeals ->
+                Log.i("Map view", "Loading meals")
```
dsmith42: Not sure whether we need to handle loading state on the map.  Whether we want to show loading spinner or some other UI?

androidHackathonApp/src/main/java/com/kcc/kmmhackathon/androidHackathonApp/view/MapsFragment.kt
```diff
@@ -36,8 +45,67 @@ class MapsFragment : Fragment() {
...
+                onLoadedMeals(state)
+            is MapsViewModel.State.LocationUpdate ->
+                onLocationUpdate(state)
+            is MapsViewModel.State.Failed ->
+                onFailure(state.failure)
+        }
+    }
+
+    private fun onLocationUpdate(state: MapsViewModel.State.LocationUpdate) {
+        map.moveCamera(CameraUpdateFactory.newLatLngZoom(state.userLatLng, 12.0f))
```
dsmith42, line 84: For now have just left the zoom as a reasonable neighbourhood.  We could consider limiting meals to a certain distance or zooming to bounds in the future.

androidHackathonApp/src/main/java/com/kcc/kmmhackathon/androidHackathonApp/viewmodel/MapsViewModel.kt
```diff
@@ -0,0 +1,96 @@
...
+
+typealias Meals = List<Meal>
+
+class MapsViewModel @ViewModelInject constructor(
+    private val fusedLocationProviderClient: FusedLocationProviderClient
+) : ViewModel() {
+
+    sealed class State {
+        object LoadingMeals : State()
+        data class LoadedMeals(val meals: Meals, val distanceUnit: DistanceUnit) : State()
```
dsmith42, line 28: Distance unit is currently unused but perhaps we could show the distance on the marker UI

feature-find-meal/src/main/java/com/kinandcarta/feature/find/meal/adapter/MealsAdapter.kt
```diff
@@ -67,7 +68,7 @@ class MealsAdapter(
             availableView.text = "Available: ${meal.availableFromDate}"
             expiryView.text = "Expires: ${meal.expiryDate}"
 
-            portionsView.text = getQuantityText(meal.quantity)
+            portionsView.text = "#    ${meal.quantity.getPortionsString()}"
```
dsmith42, line 71: Lifted the quantity portions logic up to an Int extension in the shared layer.  If desired we could just return quantity as the corrected string.

Gmiller290488: I think we need quantity for too many other things to just return the formatted string in the meal (like for whether the button is disabled, text colour etc).

With this being the shared code, we have Strings in a few separate places. In iOS we have a dedicated Strings module and so it's easy to find them if they need changing. In Android, it seems they are being set in the views (or viewModels) but then if this one bit of text is returned from an extension in the shared code it could be kind of confusing.

Maybe we should move all the Strings to shared code - and if we are going to do that then I think this is fine but otherwise I'm unsure on it

androidHackathonApp/src/main/java/com/kcc/kmmhackathon/androidHackathonApp/view/MapsFragment.kt
kirstybutler: I don't think there's usually a loading state for maps, unless you want to have a refresh. The way I've handled it previously is having the map setup as an async for getting the data in the background and once that's been fetched to add the markers onto the map, so that the user can still interact with the map like scroll or zoom while the data is being fetched. I've not personally added a spinner to a map, I think that's better for a list view. There are probably better ways to do so, though :)

androidHackathonApp/src/main/java/com/kcc/kmmhackathon/androidHackathonApp/viewmodel/MapsViewModel.kt
kirstybutler: I think this is a good idea :)

androidHackathonApp/src/main/java/com/kcc/kmmhackathon/androidHackathonApp/viewmodel/MapsViewModel.kt
```diff
@@ -0,0 +1,96 @@
...
+
+    private fun createLocationRequest(): LocationRequest {
+        val locationRequest = LocationRequest()
+        locationRequest.interval = 100000
+        locationRequest.fastestInterval = 5000
+        locationRequest.priority = LocationRequest.PRIORITY_HIGH_ACCURACY
+        return locationRequest
+    }
+
+    fun updateMeals() {
```
Gmiller290488: could it be private?

androidHackathonApp/src/main/java/com/kcc/kmmhackathon/androidHackathonApp/viewmodel/MapsViewModel.kt
```diff
@@ -0,0 +1,96 @@
...
+                sdk.getSortedMeals(location.latitude, location.longitude, distanceUnit)
+            }.onSuccess {
+                _state.value = State.LoadedMeals(it, distanceUnit)
+            }.onFailure {
+                _state.value = State.Failed(Failure.LoadingMealsFailed(it))
+            }
+        }
+    }
+
+    fun updateUserLocation() {
```
Gmiller290488: could it be private?

shared/src/commonMain/kotlin/com/kcc/kmmhackathon/shared/utility/extensions/Int+Extensions.kt
```diff
@@ -0,0 +1,7 @@
+package com.kcc.kmmhackathon.shared.utility.extensions
+
+fun Int.getPortionsString(): String = when (this) {
+    in 2 .. Int.MAX_VALUE -> "$this portions remaining"
+    1 -> "$this portion remaining"
+    else -> "Reserved"
```
Gmiller290488: I think I preferred this when it was:
```
0 -> 
1 ->
else ->
```

If we prevent uploading a meal with a negative value. And we check before we PATCH, (and I _think_ the app would crash if we actually tried to PATCH to a negative number anyway) then I think that is safe enough

androidHackathonApp/src/main/java/com/kcc/kmmhackathon/androidHackathonApp/view/MapsFragment.kt
```diff
@@ -36,8 +45,67 @@ class MapsFragment : Fragment() {
...
+        state.meals.forEach {
+            placeMarkerOnMap(LatLng(it.locationLat.toDouble(), it.locationLong.toDouble()), it.name, it.quantity.getPortionsString())
+        }
+    }
+
+    private fun placeMarkerOnMap(location: LatLng, title: String, mealsRemaining: String) {
+        val markerOptions = MarkerOptions().position(location).title(title).snippet(mealsRemaining)
+        map.addMarker(markerOptions)
+    }
+
```
kirstybutler, line 98: nice :)

feature-find-meal/src/main/java/com/kinandcarta/feature/find/meal/adapter/MealsAdapter.kt
kirstybutler: We had issues accessing the Strings file from other modules when it's in the shared. So for now we have a Strings file within each module til we find a fix - is that still the case @dsmith42 ?

shared/src/commonMain/kotlin/com/kcc/kmmhackathon/shared/utility/extensions/Int+Extensions.kt
dsmith42: 👍 Fair enough, it is a kind of impossible state for the quantity to be in.

shared/src/commonTest/kotlin/com.kcc.kmmhackathon.shared.utility/extensions/IntExtensionsTest.kt
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    @Test
+    fun testInt_getPortionsString_negativeValues() {
+        val negativeOneAvailable = -1
+        val negativeLoadsAvailable = -200
+
+        assertEquals(negativeOneAvailable.getPortionsString(), "Reserved")
+        assertEquals(negativeLoadsAvailable.getPortionsString(), "Reserved")
+    }
```
kirstybutler: Is this test necessary? Do we have the case where the portion numbers can be below 0?

kirstybutler: You can't reserve (reduce the portion # by 1) once the quantity is 0 because the button is disabled, right?

dsmith42: It was there for completeness, but as noted by both you and Gareth, there should be no case where the quantity is < 0

feature-find-meal/src/main/java/com/kinandcarta/feature/find/meal/adapter/MealsAdapter.kt
dsmith42: I agree with your thinking.  For now I can move this into an extensions  on Android while we think about the other option.

kirstybutler (approved):

androidHackathonApp/src/main/java/com/kcc/kmmhackathon/androidHackathonApp/viewmodel/MapsViewModel.kt
```diff
@@ -0,0 +1,96 @@
...
+        fusedLocationProviderClient.requestLocationUpdates(
+            createLocationRequest(),
+            locationCallback,
+            null
+        )
+    }
+
+    private fun createLocationRequest(): LocationRequest {
+        val locationRequest = LocationRequest()
+        locationRequest.interval = 100000
```
roger-tan: I would prefer to have a local variable that explains what's the value means so it can improve the readibility? Is it second, ms,?

Gmiller290488 (approved):

PR #77, opened 2020-12-10
Feature/move apikey

raynelle03:


iosHackthonApp/Modules/Features/Login/Sources/Login/LoginViewModel.swift
```diff
@@ -12,35 +12,33 @@ import Components
...
+    self.authorizationStore = authorizationStore
+    coloredNavAppearance.configureWithOpaqueBackground()
+    coloredNavAppearance.backgroundColor = UIColor(ColorManager.appPrimary)
+    coloredNavAppearance.titleTextAttributes = [.foregroundColor: UIColor.white]
+    coloredNavAppearance.largeTitleTextAttributes = [.foregroundColor: UIColor.white]
+  }
+  
+  func login(email: String, password: String) {
+    isLoading = true
+    firebase.signIn(email: email, password: password, returnSecureToken: true, completionHandler: { result, _ in
```
Gmiller290488, line 35: Just as a note (before I forget again) we could look into the security angle of the log in - so do something so that we aren't sending plain text passwords over the internet. Nothing that needs to be done here, just something we might consider picking up in the future

shared/src/commonMain/kotlin/com/kcc/kmmhackathon/shared/network/FirebaseAuthenticationApi.kt
```diff
@@ -42,7 +39,7 @@ class FirebaseAuthenticationStore : AuthenticationStore {
         return try {
             httpClient.post {
                 url("$BASE_URL$endpoint")
-                parameter("key", apiKey)
+                parameter("key", "") // add api key here
```
Gmiller290488: Would be nice to add a link to the firebase console here

iosHackthonApp/Modules/Features/Login/Sources/Login/LoginViewModel.swift
raynelle03: yup cool..will add it in our backlog 👍

shared/src/commonMain/kotlin/com/kcc/kmmhackathon/shared/network/FirebaseAuthenticationApi.kt
raynelle03: Added

Gmiller290488 (approved):

roger-tan (approved):

PR #82, opened 2020-12-14
Tidy up dependencies using buildSrc

fmontesino:
## Why?

Having hardcoded dependencies is not very useful for reusing them or updating all versions at once.

## What?

Added a very helpful Gradle plugin to keep all dependencies under `Libs` and `Versions`. 

### How to add new dependencies
1. Add your dependency as usual.
2. Run `./gradlew buildSrcVersions` and this will automatically create an entry under Libs & Versions.
3. Replace your dependency with Libs.your_dependency (Pro tip: it has autocomplete!)

###  How to check for newer versions
Just run `./gradlew buildSrcVersions` and it will automatically detect any new version for your dependencies. You'll still have to manually update the version, but it's as easy as uncommenting the newer version 🙃 

## Bonus

Ignored the `.idea` folder, we don't need/want that on git.

### It also updates the project and CI to run with Java 11.


kirstybutler (approved):
noice

common.gradle
```diff
@@ -34,41 +34,36 @@ android {
...
-    def googlePlayServicesVersion = "17.0.0"
-    def kotlinVersion = "1.4.20"
-
     // Shared multiplatform lib
     implementation project(":shared")
 
     // Kotlin
-    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-android:1.3.9"
-    implementation "org.jetbrains.kotlin:kotlin-stdlib:1.4.10"
+    implementation Libs.kotlinx_coroutines_android
```
dsmith42, line 41: Beautiful

dsmith42 (approved):
Wow, doesn't that makes things nicer to read and organise

fmontesino:
> Wow, doesn't that makes things nicer to read and organise

Wait until the next PR then @dsmith42 😉

PR #83, opened 2020-12-14
Better Gradle module organisation

fmontesino:
## Why?

Looking at the filesystem organisation, it looked like this:

```
/root
|
- androidHackatonApp
|
- feature-XXX (android module)
|
- feature-YYY (android module)
|
- iosHackatonApp
|
- lib-theming (android module)
|
- shared (multiplatform module)
```

Obviously, this is not scalable. We would end up with lots of android modules mixed with iOS and multiplatform ones.

## What?

The PR might look massive, but it's basically just reorganising the filesystem to something like this:

image

So everything is organised under its own folder 😍


kirstybutler (approved):

roger-tan (approved):
LGTM

PR #91, opened 2020-12-15
ReadMe update

### LINKED ISSUE #90: ReadMe update
Updating readme to have contribution guidelines

Gmiller290488:
I've updated the ReadMe to include a link to the projects area of the repo and with a bit of chat about the (potential) workflow.

The reason for doing this, is that I think it'll be new to most people to be using Github projects to keep a track of tickets so it's handy to point it out.

But worth mentioning that I haven't used Github projects before, so if any one has suggestions for a different way to work then I'm all ears 👂 😄 
And also that I haven't seen this automation in action, so if what I wrote in the readme turns out to be bollocks then I'll change it 😅 (using this PR as a test)

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Currently, a user can: create an account and login and search for meals in their
 ## Prerequisites
 
 ### Android Studio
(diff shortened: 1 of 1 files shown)


raynelle03 (approved):

fmontesino (approved):
Good stuff!

PR #92, opened 2020-12-15
Change title text of bottom sheet on state change 

### LINKED ISSUE #89: Change title text of bottom sheet on state change
AC's:

- [x] When the state is expanded, the title is "Hide List"
- [x] When the state is collapsed, the title is "Show List"

fmontesino:
## Why?

When the BottomSheet is collapsed we want to display "Show list", but when the BottomSheet is expanded we want to change it to "Hide list".

## What?

- Refactored the MapFragment and FindMealsFragment into one single class.
- Extracted the BottomSheet behaviour callback into its own, simplified and reusable class.
- Added some styling to the BottomSheet (background, rounded corners, etc).

## Bonus

- Tiny improvements in the codebase.
- Fixed the List section appearing in the BottomNavigation.


android/lib/theming/src/main/res/values/colors.xml
```diff
@@ -8,6 +8,7 @@
     <color name="white">#FFFFFF</color>
     <color name="black">#000000</color>
     <color name="link">#3F51B5</color>
+    <color name="bottomSheetBackground">#F8F8F8</color>
 
     //Meal
```
Gmiller290488: Is it worth changing this comment over to the `<!-- -->` style also? (I guess the // isn't standard for Android?)

Gmiller290488 (approved):

android/lib/theming/src/main/res/values/colors.xml
fmontesino: Very true! Will fix 😄 

Yeah, `//` comments on XML is definitely not a standard

android/feature/find-meal/src/main/java/com/kinandcarta/feature/find/meal/view/FindMealFragment.kt
```diff
@@ -1,67 +1,113 @@
...
 @AndroidEntryPoint
 class FindMealFragment : Fragment() {
 
-    private val progressBarView: FrameLayout by lazy { binding.progressBar }
     private val viewModel: DisplayMealsViewModel by viewModels()
-    private val mealsRecyclerView: RecyclerView by lazy { binding.rvMeals }
-    private val mealsAdapter = MealsAdapter { id -> viewModel.reserveAMeal(id) }
+
     private var _binding: FindMealFragmentBinding? = null
     private val binding get() = _binding!!
```
roger-tan, line 32: What does it mean the double question mark for a Kotlin point of view? As a Swift developer, I feel very scary to see   this.

fmontesino: It literally means: this nullable variable _**won't**_ be `null` at the time it's accessed. 

As a Kotlin developer, I am also very afraid of seeing `!!` hahaha. As a general rule of thumb, I'd decline any PR containing `!!`, but this case is well justified, let me explain why:

When a Fragment calls `onCreateView` we use ViewBinding to get the layout, but when the Fragment calls `onDestroyView` we **_must_** destroy that layout reference. (This is one of the changes a few lines below this).
In other words, we can guarantee that `_binding` won't be null between `onCreateView` and `onDestroyView`.

IMO it's a bit ugly, but it's the only recommended way of dealing with `ViewBinding` according to the [documentation](https://developer.android.com/topic/libraries/view-binding#fragments).

roger-tan (approved):
LGTM

PR #93, opened 2020-12-16
Move Add a meal to a new activity

roger-tan:
What I've been done :
- Removed some old fragment from the navhost.
- Created a container of activities for the maps fragment and add a meal fragment as `feature-add-meal` and `feature-find-meal` should not know each other. Let me know about this. 

Quick preview: 
ezgif com-resize

### DIFF
--- a/android/app/src/main/java/com/kcc/kmmhackathon/androidHackathonApp/view/FindMealContainerFragment.kt
+++ b/android/app/src/main/java/com/kcc/kmmhackathon/androidHackathonApp/view/FindMealContainerFragment.kt
@@ -1,0 +1,54 @@
+package com.kcc.kmmhackathon.androidHackathonApp.view
+
+import android.content.Intent
+import android.os.Bundle
+import android.view.*
+import androidx.fragment.app.Fragment
+import com.kcc.kmmhackathon.androidHackathonApp.R
+import com.kinandcarta.lib.add.meal.view.AddMealFragment
+import kotlinx.coroutines.Dispatchers.Main
+
+class FindMealContainerFragment : Fragment() {
+
+    companion object {
+        fun newInstance() = FindMealContainerFragment()
+    }
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setHasOptionsMenu(true)
+    }
+
+    override fun onCreateView(
+        inflater: LayoutInflater, container: ViewGroup?,
+        savedInstanceState: Bundle?
+    ): View? {
+        val view = inflater.inflate(R.layout.find_meal_container_fragment, container, false)
+        val fragment = com.kinandcarta.feature.find.meal.view.MapsFragment()
+        childFragmentManager.beginTransaction().apply {
+            add(R.id.child_fragment_container, fragment)
+            commit()
+        }
+        return view
+    }
+
+    override fun onCreateOptionsMenu(menu: Menu, inflater: MenuInflater) {
+        super.onCreateOptionsMenu(menu, inflater)
+
+        inflater.inflate(R.menu.find_meal_container_menu, menu)
+    }
+
+    override fun onOptionsItemSelected(item: MenuItem): Boolean {
+        return when (item.itemId) {
+            R.id.add_an_item -> {
+                val intent = Intent(activity, AddMealActivity::class.java)
+                startActivity(intent)
+                true
+            }
+            else -> {
+                super.onOptionsItemSelected(item)
+            }
+        }
+    }
+
+}
--- a/android/app/src/main/res/menu/bottom_nav_menu.xml
+++ b/android/app/src/main/res/menu/bottom_nav_menu.xml
@@ -3,18 +3,9 @@
     xmlns:app="http://schemas.android.com/apk/res-auto">
 
     <item
-        android:id="@+id/mapsFragment"
+        android:id="@+id/findMealFragment"
         android:icon="@drawable/ic_baseline_map_24"
         android:title="@string/map_tab_label" />
-
-    <item
-        android:id="@+id/addMealFragment"
-        android:icon="@drawable/ic_baseline_add_24"
-        android:title="@string/add_meal_tab_label" />
-
-    <item android:id="@+id/findMealFragment"
-        android:icon="@drawable/ic_baseline_list_24"
-        android:title="@string/title_switch_to_list"/>
 
     <item
         android:id="@+id/settingsFragment"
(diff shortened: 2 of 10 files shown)


android/app/src/main/java/com/kcc/kmmhackathon/androidHackathonApp/view/FindMealContainerFragment.kt
```diff
@@ -0,0 +1,54 @@
...
+            commit()
+        }
+        return view
+    }
+
+    override fun onCreateOptionsMenu(menu: Menu, inflater: MenuInflater) {
+        super.onCreateOptionsMenu(menu, inflater)
+
+        inflater.inflate(R.menu.find_meal_container_menu, menu)
+    }
```
kirstybutler, line 39: I think we're having the adding of a meal within a second tab of the bottom nav rather than this

android/app/src/main/java/com/kcc/kmmhackathon/androidHackathonApp/view/FindMealContainerFragment.kt
```diff
@@ -0,0 +1,54 @@
+package com.kcc.kmmhackathon.androidHackathonApp.view
```
kirstybutler, line 1: I'm confused as to why you've created this file - please could you give some more context?

android/app/src/main/res/menu/bottom_nav_menu.xml
```diff
@@ -3,19 +3,10 @@
...
-        android:icon="@drawable/ic_baseline_add_24"
-        android:title="@string/add_meal_tab_label" />
-
-    <item android:id="@+id/findMealFragment"
-        android:icon="@drawable/ic_baseline_list_24"
-        android:title="@string/title_switch_to_list"/>
-
     <item
         android:id="@+id/settingsFragment"
         android:icon="@drawable/ic_baseline_settings_24"
```
kirstybutler, line 12: This is going to conflict with Fidel's PR with the refactoring of the `FindMealFragment`, no?

android/app/src/main/res/menu/find_meal_container_menu.xml
```diff
@@ -0,0 +1,8 @@
+<menu xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:yourapp="http://schemas.android.com/apk/res-auto" >
+
+    <item android:id="@+id/add_an_item"
+        android:title="Add"
+        android:orderInCategory="100"
+        yourapp:showAsAction="always" />
+</menu>
```
kirstybutler, line 8: Again I think this should stay as a tab :)

android/app/src/main/res/values/strings.xml
```diff
@@ -23,7 +23,7 @@
     <string name="hint_password">Enter Password</string>
 
     //Headers
-    <string name="add_meal_activity_header">This is the add meal activity screen</string>
```
kirstybutler, line 26: Do we need it as an Activity?

android/feature/find-meal/src/main/java/com/kinandcarta/feature/find/meal/viewmodel/DisplayMealsViewModel.kt
```diff
@@ -55,9 +55,11 @@ class DisplayMealsViewModel @ViewModelInject constructor(
     @RequiresPermission("android.permission.ACCESS_FINE_LOCATION")
     fun startUpdatingLocation() {
         fusedLocationProviderClient.lastLocation.addOnSuccessListener {
-            val latLng = LatLng(it.latitude, it.longitude)
-            updateUserLocation(latLng)
-            updateMeals(latLng)
+            if (it != null) {
```
kirstybutler, line 58: Is this null check necessary? Would we get to `startUpdatingLocation` if there was no current location?

fmontesino (changes requested):
MapsFragment doesn't exist anymore 😕 it will definitely conflict with the open PR.
Also, as a general Android UX, having the Add button on the Toolbar is very confusing, I'd rather create a FAB  on the Find A Meal fragment that opens the Add screen.

kirstybutler (commented):
Also, just as a minor thing, when creating a PR please could you link it to an issue within the Projects? And also add a label as `Android` so that it's easy to keep track of which platform this is for? It will help make it easier for new people joining the project to understand more of what's going on and also for us to view the board in a more organised way 🙌  😃

android/app/src/main/java/com/kcc/kmmhackathon/androidHackathonApp/view/FindMealContainerFragment.kt
roger-tan: I created this file because I don't think that the find MapsFragment should know when to create a new add meal activity

android/app/src/main/java/com/kcc/kmmhackathon/androidHackathonApp/view/FindMealContainerFragment.kt
roger-tan: In the iOS side, it was a new modal from the map.

android/app/src/main/res/menu/bottom_nav_menu.xml
roger-tan: Pretty sure that's going to be conflicted.

android/app/src/main/res/values/strings.xml
roger-tan: If we want to have the toolbar yes!

android/feature/find-meal/src/main/java/com/kinandcarta/feature/find/meal/viewmodel/DisplayMealsViewModel.kt
roger-tan: It was crashing on my side without checking if it's null.

android/app/src/main/java/com/kcc/kmmhackathon/androidHackathonApp/view/FindMealContainerFragment.kt
fmontesino: Good point! But following that logic, the `FindMeal` also shouldn't know about creating new meals and creating a new container just for that seems a bit overkill.

I'd add this logic in the Activity holding the Fragments and BottomNavigation, maybe in a FloatingActionButton.

Either way, FAB or action in the Toolbar, the action belongs to the Activity, and then you have 3 independent Fragments to navigate to.

PR #95, opened 2020-12-17
Refactor permission using a PermissionManager class

### LINKED ISSUE #96: Refactor permissions using a PermissionManager class

fmontesino:
## Why?

We had some Fragment extension functions very specific for location permissions. Although it worked, it wasn't elegant and we could extract that functionality into a clean, reusable class.

## What?

Created a new class called `PermissionManager` that will register into any Fragment to listen to Permission results.

Abstracted the Permission request and whole permissions decision tree (based on the official Android best practices).

In my eyes, it reads easier and cleaner. But if you have any question please shout!


android/lib/core/src/main/java/com/kinandcarta/kmmhackathon/lib/core/permissions/PermissionManager.kt
```diff
@@ -0,0 +1,109 @@
...
+            .setMessage(rationale ?: fragment.getString(R.string.dialog_permission_default_message))
+            .setCancelable(false)
+            .setPositiveButton(fragment.getString(R.string.dialog_permission_button_ok)) { _, _ ->
+                requestPermissions()
+            }
+            .show()
+    }
+
+    private fun sendPositiveResult() {
+        sendResultAndCleanUp(getPermissionList().associate { it to true })
```
kirstybutler, line 69: niceeee

buildSrc/src/main/kotlin/Versions.kt
```diff
@@ -56,7 +56,7 @@ object Versions {
 
     const val constraintlayout: String = "2.0.4"
 
-    const val fragment_ktx: String = "1.2.5"
+    const val fragment_ktx: String = "1.3.0-beta02"
```
kirstybutler, line 59: I was going to say to mention why this has changed, but maybe not necessary 🤷

kirstybutler (approved):

buildSrc/src/main/kotlin/Versions.kt
fmontesino: Well, just in case:

The functionality used in the PermissionManager for registering the fragment to the permission result is only in the beta version of the library

Gmiller290488 (approved):
Looks good 👍

PR #97, opened 2020-12-21
Feature/signout

### LINKED ISSUE #86: Add SignOut
AC's:

- [x] User is diverted back to login screen
- [x] User session is stopped and there is no current user with Firebase Auth
- [x] The UI for the SignOut functionality within the Profile is complete

raynelle03:


raynelle03:
Just explaining how the pop to root functionality works in swiftUI.

We are using an environment object variable `moveToHome` to keep track of when the views should be popped to the root view `LoginView`. By default this value is false.

We are listening to this variable on the root view i.e `LoginView` and are setting it to true on the `SettingsView` when the user taps the `Sign out` button.

So on the `LoginView` when this value is received as true, we then set the binding values on then NavigationLink for `isActive` to `false`, and clear the authorisation store.  This causes the views to pop back to the root view I.e : `LoginView`.


Another important thing to note is setting the `.isDetailLink(false)` modifier on the NavigationLink. This will ensure that the destination view will always be pushed onto the navigation stack and can always be popped off.

ios/iosHackthonApp/Scene Delegate/SceneDelegate.swift
```diff
@@ -11,9 +11,13 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
...
 
     let loginView = LoginView(viewModel: loginViewModel)
+    let appState = AppState()
+
     if let windowScene = scene as? UIWindowScene {
       let window = UIWindow(windowScene: windowScene)
-      window.rootViewController = UIHostingController(rootView: loginView)
+      window.rootViewController = UIHostingController(rootView:
+                                                        loginView
+                                                        .environmentObject(appState))
```
Gmiller290488: This looks a little strange to me. I don't know if it is possible but could we do the `.environmentObject` on line 13?

ios/iosHackthonApp/Scene Delegate/SceneDelegate.swift
```diff
@@ -11,9 +11,13 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
   func scene(_ scene: UIScene, willConnectTo session: UISceneSession, options connectionOptions: UIScene.ConnectionOptions) {
 
     let loginView = LoginView(viewModel: loginViewModel)
+    let appState = AppState()
```
Gmiller290488: if we switch these two lines, could we then do
`let loginView = LoginView(viewmodel: loginViewModel).environmentObject(appState)`

or 
```
let loginView = LoginView(viewmodel: loginViewModel)
loginView.environmentObject(appState)
```

Something like this?

ios/iosHackthonApp/Screens/Login/LoginView.swift
```diff
@@ -97,5 +103,27 @@ struct LoginView: View {
...
+    self.appState.moveToHome = false
+    self.isSignUpButtonPressed = false
+    self.loginViewModel.authorizationStore.isAuthorised = false
+    self.loginViewModel.clear()
+    self.resetEmailAndPassword()
+  }
+
+  func resetEmailAndPassword() {
+    self.loginViewModel.email = ""
+    self.loginViewModel.password = ""
```
Gmiller290488: do we need all the self's?

ios/iosHackthonApp/Scene Delegate/SceneDelegate.swift
raynelle03: yup..changed to 
`let appState = AppState()`
` let loginView = LoginView(viewModel: loginViewModel).environmentObject(appState)`

ios/iosHackthonApp/Screens/Login/LoginView.swift
raynelle03: Removed

ios/iosHackthonApp/Scene Delegate/SceneDelegate.swift
raynelle03: Yes changed this.

Gmiller290488 (approved):

ios/iosHackthonApp/Screens/Login/LoginView.swift
```diff
@@ -97,5 +103,27 @@ struct LoginView: View {
...
+
+// MARK: - Functions
+private extension LoginView {
+
+  func resetOnLogOut() {
+    appState.moveToHome = false
+    isSignUpButtonPressed = false
+    loginViewModel.authorizationStore.isAuthorised = false
+    loginViewModel.clear()
+    resetEmailAndPassword()
```
fmontesino: This whole `resetOnLogout` it's doing too much for what the name says, IMO.

It's clearing some states, logging out the user, resetting data and also triggering the navigation.

Maybe it's me being completely new to Swift/iOS, but I'd find this a bit confusing to follow.
Something like you did with `resetEmailAndPassword()` where you bundle some small functionality would help. For example: `logoutUserAndNavigateHome()` and `resetUI()` ?

ios/Modules/Components/Sources/Components/States/AppState.swift
```diff
@@ -0,0 +1,16 @@
...
+//
+//  Created by Raynelle Francisca on 18/12/2020.
+//  Copyright © 2020 orgName. All rights reserved.
+//
+
+import Foundation
+import Combine
+
+public class AppState: ObservableObject {
+  @Published public var moveToHome: Bool = false
```
fmontesino: This feels a bit hacky to me. Sure it works now, but when you have more complex navigations or you just have multiple navigations, you'll end up with a ton of `Bool` flags that you need to take care of every single time.

What about making this a custom class or an enum, something like:
```
public var navDestination: NavDestination? = nil // make it nil as default!
```

Then your NavDestination enum could me something like:
```
enum NavDestination{ 
Home, SectionA, SectionB, SectionC // Add more in the future
}
```

Finally, your view would receive this navDestination like:
```
.onReceive(self.appState.$navDestination) { navDestination in
    // if navDestination is nil, do nothing
    // check here if navDestination is Home, then navigate to home
    // else, check if navDestination is SectionA, then navigate accordingly
    // etc
    // reset navDestination to nil
}
```

I don't know if this would be a good approach for an iOS app, but that definitely would centralise your navigation and it takes care of reusability, scalability and testability.

fmontesino (approved):
I'm approving this because I trust @Gmiller290488's approval hahaha
I just left a couple of comments on how I would like to approach this, as an engineer, but maybe they don't follow the iOS conventions so feel free to ignore

ios/Modules/Components/Sources/Components/States/AppState.swift
Gmiller290488: I think this sounds like a nice approach.
It would work in a similar way to how we handle the alerts in the app @raynelle03

raynelle03: Umm..tbh not sure how this would work in swiftUI. 😅
I have followed the approach that’s currently out there.
The whole purpose of `moveToHome`  is to basically move to the RootView.

`moveToHome` is an environment object and can be set to true on any view. 

Currently we have 2 flows for login and signup

Login:     LoginView -> HomeScreen->SettingsView
SignUp:  LoginView -> RegisterWithEmailScreen -> EnterTheEmailScreen ->EnterThePasswordScreen ->  HomeScreen-> SettingsView


The `moveToHome` will only be observed on the RootView i.e : `LoginView`. And the only place it would be set , is on the  `SettingsView` When the user taps `Logout button` (So you set it to true only if u want to go back to the root view).
And then accordingly the LoginView navigationLinks `isActive` is set to false. Causing the pop to root view. 

It works for both the flows as we’re only observing it on the RootView (`LoginView`) and setting it on the views we want to pop to the root view.

Even if we have a complex flow(multiple screens), it will still work, as we will only need to set this `moveToHome` to true. and the views should react to this and pop accordingly.

ios/iosHackthonApp/Screens/Login/LoginView.swift
raynelle03: yup agreed..I ll try and simplify this

ios/Modules/Components/Sources/Components/States/AppState.swift
fmontesino: @raynelle03 the idea is to keep the same approach that you applied here, but instead of using a `Bool` to decide where you want to go (and then you have to turn that `Bool` to false!), you have a generic Destination object.

My fear is that you might end up with a lot of states like:
`moveToHome`, `moveToDetails`, `moveToSettings`, etc... 
and then, there's the possibility to leave one of those flags to `true` mistakenly and that will produce some weirdness/bugs in the app.

However, if you only have one object that could either be Home, Details or Settings, there's no state to keep/reset and you minimise the probability of bug, minimise the code/states maintenance and you don't have to change your AppState class every time you want a new navigation

Gmiller290488: Navigation in SwiftUI is also something I've not had to handle outside of just using NavigationLinks to move to a screen and back again so I appreciate your concerns.

raynelle03: Ahhh okay..think I understand now

raynelle03: Okay done...nice one @fmontesino... I din't understand at first😅. But yeah this approach is more scalable!
Added `NavDestination` to appState

ios/iosHackthonApp/Screens/Login/LoginView.swift
raynelle03: broke this to `resetAuthorizationStoreAndLogout` and `resetEmailAndPassword`

fmontesino (approved):
Good one!

